PR #1, opened 2024-08-13
Update metadata endpoints response types

katia-qm:
**Background**
Metadata creation endpoints have changed their response types from None.

**Changes**
- created response types for metadata endpoints
- updates parser to work with parametrized generics


hari_client/client/client.py
```diff
@@ -1273,5 +1318,7 @@ def create_crops(
             f"/datasets/{dataset_id}/crops",
             params={"subset_id": subset_id},
             json=self._pack(locals(), ignore=["dataset_id", "subset_id"]),
-            success_response_item_model=None,
+            success_response_item_model=list[
+                models.UpdateHistogramsResponse | models.CreateCropsResponse
```
Dens49, line 1322: our minimum python version is 3.9, but using the pipe symbol in the typehint isn't supported in python 3.9. Can you switch to using typing.Union?

Dens49 (changes requested):
Only 2 change requests:
- minimum python version we support is 3.9, but using the pipe for a union typehint isn't supported in python 3.9. All the new models have to be updated
- can you explain the support of the union in the docstring of `_parse_response_model` along with its limitations? For example the limits on nesting

hari_client/client/client.py
```diff
@@ -1312,13 +1325,18 @@ def create_crops(
 
         Raises:
             APIException: If the request fails.
+
+        Returns:
+            list[models.UpdateHistogramsResponse, models.CreateCropsResponse]: list of updated histograms and created crops
```
Dens49, line 1330:
```suggestion
            list[typing.Union[models.UpdateHistogramsResponse, models.CreateCropsResponse]]: list of updated histograms and created crops
```

katia-qm: oops, thanks for finding this!

Dens49 (approved):
looks good! just a tiny update for one of the docstrings

PR #3, opened 2024-08-14
Km/wrap processing job endpoints

katia-qm:
**Background**
Create wrappers around endpoints for processing jobs

### DIFF
--- a/hari_client/client/client.py
+++ b/hari_client/client/client.py
@@ -1340,3 +1340,31 @@ class HARIClient:
                 ],
             ],
         )
+
+    def get_processing_jobs(
+        self,
+        trace_id: str = None,
+    ) -> list[models.ProcessingJob]:
+        """
+        Retrieves the list of processing jobs that the user has access to.
+
+        Args:
+            trace_id (str, optional): A unique identifier of the trace. Defaults to None.
+
+        Raises:
+            APIException: If the request fails.
+
+        Returns:
+            list[models.ProcessingJob]: A list of processing jobs for the user
+            or [] if there are no jobs of trace_id is not found.
+        """
+        params = {}
+        if trace_id:
+            params["trace_id"] = trace_id
+
+        return self._request(
+            "GET",
+            "/processingJobs/",
+            params=params,
+            success_response_item_model=list[models.ProcessingJob],
+        )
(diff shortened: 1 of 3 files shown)


hari_client/client/client.py
```diff
@@ -1339,3 +1339,54 @@ def create_crops(
                 ],
             ],
         )
+
```
Dens49, line 1342: nitpick: can you add the name of the endpoint group as a comment here? you can see that group name in the openapi documentation site. This helps with keeping a bit of an overview over the many methods here.

hari_client/models/models.py
```diff
@@ -918,3 +918,16 @@ class CreateCropsResponse(ResponseBaseParameters):
...
+    id: str = pydantic.Field(title="ID")
+    status: str = pydantic.Field(title="Status")
+    owner: str = pydantic.Field(default=None, title="Owner")
+    user_group: str = pydantic.Field(default=None, title="User Group")
+    created_at: str = pydantic.Field(title="Created At")
+    updated_at: str = pydantic.Field(title="Updated At")
+    archived_at: typing.Optional[str] = pydantic.Field(title="Archived At")
+    process_name: str = pydantic.Field(title="Process Name")
+    details: str = pydantic.Field(title="Details")
+    trace_id: typing.Optional[str] = pydantic.Field(title="Trace ID")
```
Dens49, line 933: the schema doesn't seem right. For example created_at, updated_at, owner are all optional. Can you check again please?

Dens49 (changes requested):
The ProcessingJob model doesn't seem right :grimacing:

hari_client/client/client.py
```diff
@@ -1348,7 +1370,7 @@ def get_processing_jobs(
         Retrieves the list of processing jobs that the user has access to.
 
         Args:
-            trace_id (str, optional): A unique identifier of the trace. Defaults to None.
+            trace_id (str, optional): A trace id to identify related processing jobs
```
Dens49, line 1373:
```suggestion
            trace_id (str, optional): Helps to identify related processing jobs. Use the trace_id that was specified when triggering a processing job
```

Dens49 (approved):
thanks for fixing! just one small suggestion for the docstring

PR #4, opened 2024-08-14
Package maintenance and pytest workflow

Dens49:
- Changed minimum python version to 3.11, due to the features used in https://github.com/quality-match/hari-client/pull/1
- Added github actions workflow for running pytest tests when creating a PR towards `main`
- Link Changelog in Readme
- Explain development setup in Readme

### DIFF
--- a/.github/workflows/hari-client-pytest.yaml
+++ b/.github/workflows/hari-client-pytest.yaml
@@ -27,5 +27,16 @@ jobs:
           python -m pip install --upgrade pip
           python -m pip install -e .[tests]
 
+      - name: Get changed files
+        id: changes
+        run: echo "files=$(git diff --name-only origin/master...${{ github.event.pull_request.head.sha }} | xargs)" >> $GITHUB_OUTPUT
+
+      - name: Run pre-commit on changed files
+        run: |
+          changed_files="${{ steps.changes.outputs.files }}"
+          if [ -n "$changed_files" ]; then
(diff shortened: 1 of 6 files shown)


.github/workflows/hari-client-pytest.yaml
```diff
@@ -0,0 +1,31 @@
...
+    branches:
+      - main
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        python-version: [3.11, 3.12]
```
katia-qm, line 14: How will this be maintained in the future with new versions?

katia-qm (approved):
Kewl! I only don't really get, why pre-commit-test workflow is necessary. As long as developers don't skip it, it will act exactly like the local one, right?

Dens49:
> Kewl! I only don't really get, why pre-commit-test workflow is necessary. As long as developers don't skip it, it will act exactly like the local one, right?

IMO It's the same reasoning as for having automated tests in the PR: you should have them running locally, but the automated checks in the PR make it visible to the reviewer that everything passed.

.github/workflows/hari-client-pytest.yaml
Dens49: When a new python version is released, we can just add it to the matrix. If tests for the new version fail, we know we have to do some changes before being able to support the new version. Right now, the setup.py file states that the package is working with python 3.11 and 3.12

PR #6, opened 2024-08-20
Update example with status tracking

katia-qm:
Update quickstart script with checking thumbnails creation status before triggering crops creation

### DIFF
--- a/docs/example_code/quickstart.py
+++ b/docs/example_code/quickstart.py
@@ -1,3 +1,6 @@
+import time
+import uuid
+
 from hari_client import Config
 from hari_client import HARIClient
 from hari_client import models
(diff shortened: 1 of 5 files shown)


docs/example_code/quickstart.py
```diff
@@ -51,10 +54,33 @@
...
+    hari.trigger_histograms_update(
+        new_dataset.id, compute_for_all_subsets=True, trace_id=trace_id
+    )
+
+    # in order to trigger crops creation, thumbnails should be created first.
+    # query all the jobs for the given trace_id
+    jobs = hari.get_processing_jobs(trace_id=trace_id)
+    # get the thumbnails job id
+    thumbnails_job_id = next(
+        (job.id for job in jobs if job.process_name == "create_thumbnails"), ""
```
Dens49, line 74: where can you find the `job.process_name`?

Dens49, line 74: is there a way to skip these lines and only use the trace_id to get the status?

docs/example_code/quickstart.py
```diff
@@ -51,10 +54,33 @@
...
+    )
+
+    job_status = ""
+    while job_status != "success":
```
Dens49, line 78: nitpick: is there an enum you can use for the status check?

hari_client/client/client.py
```diff
@@ -1376,7 +1376,7 @@ def add_visualisation_to_media_object(
         )
 
     ### metadata ###
-    def create_thumbnails(
+    def trigger_thumbnails_creation(
```
Dens49, line 1379: nitpick: maybe even `_job` as a suffix for all of them? --> `trigger_thumbnails_creation_job`?

Dens49 (changes requested):
Is there a way to get the processing job status just from the trace_id, without having to find the thumbnails_job_id first?

docs/example_code/quickstart.py
katia-qm: created a enum for it

katia-qm: Nope, for get processing job endpoint we could only filter by job_id.
We can filter by trace id in get processing jobS endpoint, and then we need find which of the jobs is for thumbnails creation

katia-qm: Nope, for `get processing job` endpoint we could only filter by job_id.
We can filter by trace id in `get processing jobS` endpoint and there is actually the job status for each job in the response, but in case if the status is not success, we will need to call this endpoint to retrieve all the jobs and loop over them to find the needed one again and again

Dens49: ah ok thanks for the explanation :+1:

Dens49 (approved):
thanks!

PR #11, opened 2024-08-28
Use proper pydantic models for response parsing

Schmaexxi:
# Background
Response parser is quite complex. Returning lists, unions or other custom types is confusing.

# Changes
- [Allow extra attributes in responses](https://github.com/quality-match/hari-client/commit/3cc6ec7eb765b43b4f88f3832fa195ee11bb4dcf)
  - added this to any type that looked like it was used for parsing responses to add this information for the user. So far, it was ignored
- [Add models for endpoints returning lists](https://github.com/quality-match/hari-client/commit/13ddb514cfe0f5ef9511268523a8427b281c87fb)
  - add models inheriting from pydantic.Rootmodel to make response parsing easier
- [Update parsing of response models](https://github.com/quality-match/hari-client/commit/b6182b447efb3786315acc93902ea7770f5910ea)
  - make response parsing easier
- [Update create_subset-method to reflect API](https://github.com/quality-match/hari-client/pull/11/commits/0f8a87d05ed7c3d5450cf442718d95ac082b1512)

### DIFF
--- a/hari_client/client/client.py
+++ b/hari_client/client/client.py
@@ -24,15 +24,8 @@ def _parse_response_model(
         - both response_data and response_model are None (meaning you expect to receive None as response)
             - None is returned
         - response_model is a pydantic model:
-            - if response_data is a dict, response_data is parsed into an instance of the response_model.
-            - if response_data is a list, each item in the list is treated as a dict and parsed into an instance of the response_model.
-        - response_model is a parametrized generic:
-            - if response_data is a list and response_model is a list of unions:
-                - each item in the response_data list is checked against the possible types in the union.
-                - the parser attempts to parse each item using each type in the union until one succeeds.
-                - if an item is successfully parsed into one of the union types, and it contains nested fields that are also complex structures (e.g., lists of unions), the parser recurses into those nested structures to fully parse them.
-            - if response_data is a dict and response_model is a dict with specified key and value types:
-                - each key-value pair in the response_data dict is parsed, with the keys being converted to the specified key type, and the values being recursively parsed according to the specified value type.
+            - if response_data is a dict or list, response_data is parsed into an instance of the response_model.
+            - raises error otherwise
         - response_data is of the expected type (response_model):
             - The response_data is returned as is.
 
@@ -44,86 +37,52 @@ def _parse_response_model(
         errors.ParseResponseModelError: When parsing fails for any reason
 
     Returns:
-        T: The passed response_model type
+        T: an instance of the passed response_model type
     """
-    # The response_data can have many different types:
-    # --> custom classes, dict, list, primitives (str, int, etc.), None
     try:
-        if response_model is None:
-            if response_data is None:
+        if response_model is None or response_data is None:
+            if response_model is None and response_data is None:
                 return None
             raise errors.ParseResponseModelError(
                 response_data=response_data,
                 response_model=response_model,
-                message=f"Expected response_data to be None, but received {response_data=}",
+                message=f"Failed to parse response_data into response_model {response_model}. {response_data=}. Types do not match.",
             )
-
         # handle pydantic models
+        # order is important here, because the RootModel is a subclass of the BaseModel
         if isinstance(response_model, type) and issubclass(
+            response_model, pydantic.RootModel
+        ):
+            root_field_type = typing.get_origin(
+                response_model.model_fields["root"].annotation
+            )
+            if issubclass(root_field_type, list):
+                return response_model(response_data)
+            else:
+                raise errors.ParseResponseModelError(
+                    response_data=response_data,
+                    response_model=response_model,
+                    message=f"Failed to parse response_data into response_model {response_model}. {response_data=}, because the root field type of the response_data is '{root_field_type}', which is not supported.",
+                )
+        elif isinstance(response_model, type) and issubclass(
             response_model, pydantic.BaseModel
         ):
-            if isinstance(response_data, dict):
-                return response_model(**response_data)
-            elif isinstance(response_data, list):
-                return [response_model(**item) for item in response_data]
-
-        # handle parametrized generics
-        origin = typing.get_origin(response_model)
-        if origin is list:
-            item_type = typing.get_args(response_model)[0]
-            if isinstance(response_data, list):
-                if typing.get_origin(item_type) is typing.Union:
-                    return [
-                        handle_union_parsing(item, item_type) for item in response_data
-                    ]
-                elif isinstance(item_type, type) and issubclass(
-                    item_type, pydantic.BaseModel
-                ):
-                    return [item_type(**item) for item in response_data]
-                else:
-                    return [item_type(item) for item in response_data]
-        if origin is dict:
-            key_type, value_type = typing.get_args(response_model)
-            if isinstance(response_data, dict):
-                return {
-                    key_type(k): (
-                        value_type(**v) if isinstance(v, dict) else value_type(v)
-                    )
-                    for k, v in response_data.items()
-                }
+            return response_model(**response_data)
 
         if isinstance(response_data, response_model):
             return response_data
-
-        raise errors.ParseResponseModelError(
-            response_data=response_data,
-            response_model=response_model,
-            message=f"Can't parse response_data into response_model {response_model},"
-            + f" because the combination of received data and expected response_model is unhandled."
-            + f"{response_data=}.",
-        )
-    except Exception as err:
+        else:
+            raise errors.ParseResponseModelError(
+                response_data=response_data,
+                response_model=response_model,
+                message=f"Failed to parse response_data into response_model {response_model}. {response_data=}. Types do not match.",
+            )
+    except pydantic.ValidationError as err:
         raise errors.ParseResponseModelError(
             response_data=response_data,
             response_model=response_model,
             message=f"Failed to parse response_data into response_model {response_model}. {response_data=}",
         ) from err
-
-
-def handle_union_parsing(item, union_type):
-    for possible_type in typing.get_args(union_type):
-        if isinstance(possible_type, type) and issubclass(
-            possible_type, pydantic.BaseModel
-        ):
-            try:
-                return possible_type(**item)
-            except Exception:
-                continue
-    raise errors.ParseResponseModelError(
-        response_data=item,
-        response_model=union_type,
-        message=f"Failed to parse item into one of the union types {union_type}. {item=}",
-    )
 
 
 class HARIClient:
@@ -143,7 +102,7 @@ class HARIClient:
         url: str,
         success_response_item_model: typing.Type[T],
         **kwargs,
-    ) -> typing.Union[T, None]:
+    ) -> T:
         """Make a request to the API.
 
         Args:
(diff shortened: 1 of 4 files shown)


hari_client/models/models.py
```diff
@@ -8,6 +8,10 @@
 import pydantic
 
 
+class BaseResponse(pydantic.BaseModel):
+    model_config = pydantic.ConfigDict(extra="allow")
+
+
```
katia-qm, line 14: I clarified with Matthias about the error he had, when the client broke with extra fields for get_datasets endpoint. However the problem was on his side, the client handles it correctly and just ignores extra field without breaking. But the test you wrote surprisingly breaks :)

tests/test_parse_response_model.py
```diff
@@ -14,12 +14,22 @@ class SimpleModel1(pydantic.BaseModel):
...
+
+
 class SimpleModel2(pydantic.BaseModel):
     x: bool
     y: list
     z: dict
 
 
+class SimpleModelWithRootFieldForUnionOfTwoModels(
+    pydantic.RootModel[list[SimpleModel1 | SimpleModel2]]
```
katia-qm, line 28: Should the union be written as a pipe or typing.Union? I think now it's inconsistent throughout the code

tests/test_parse_response_model.py
```diff
@@ -81,122 +91,88 @@ def test_parse_response_model_works_with_list_of_pydantic_models():
...
-    assert isinstance(response, list)
-
     for idx, item in enumerate(response):
         assert isinstance(item, models.MediaUploadUrlInfo)
         assert item.upload_url == response_data[idx]["upload_url"]
         assert item.media_id == response_data[idx]["media_id"]
         assert item.media_url == response_data[idx]["media_url"]
 
 
+type_mismatch_error_match = ".Types do not match."
```
katia-qm, line 104: Nitpick: will work without the dot in the beginning

hari_client/models/models.py
```diff
@@ -297,6 +301,14 @@ class DatasetResponse(pydantic.BaseModel):
     )
 
 
+class DatasetResponseList(pydantic.RootModel[list[DatasetResponse]]):
+    def __iter__(self):
+        return iter(self.root)
+
+    def __getitem__(self, item):
+        return self.root[item]
```
katia-qm, line 309: It looks to complicated to me, I think having list[Model] looks way more readable. Do you think it's too hard to update the parser to be able to parse plain lists?

katia-qm (commented):
Thanks for the effort and your clean-ups!

I have some questions about the necessity of the complexity there(imho).

hari_client/models/models.py
Schmaexxi: Which test breaks? Pytest passes all tests for me 🤔 
I think a client should not break because an API returns more information than you expect. This is exactly what we discussed and defines what a breaking change for an API is.
Adding information to a response is not a breaking change. Altering existing response model is

hari_client/client/client.py
```diff
@@ -44,88 +37,54 @@ def _parse_response_model(
...
     """
-    # The response_data can have many different types:
-    # --> custom classes, dict, list, primitives (str, int, etc.), None
     try:
-        if response_model is None:
-            if response_data is None:
+        if response_model is None or response_data is None:
+            if response_model is None and response_data is None:
                 return None
             raise errors.ParseResponseModelError(
```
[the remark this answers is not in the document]
Schmaexxi: looks more intuitive, thanks!

hari_client/client/client.py
```diff
@@ -24,15 +24,8 @@ def _parse_response_model(
...
-            - if response_data is a list, each item in the list is treated as a dict and parsed into an instance of the response_model.
-        - response_model is a parametrized generic:
-            - if response_data is a list and response_model is a list of unions:
-                - each item in the response_data list is checked against the possible types in the union.
-                - the parser attempts to parse each item using each type in the union until one succeeds.
-                - if an item is successfully parsed into one of the union types, and it contains nested fields that are also complex structures (e.g., lists of unions), the parser recurses into those nested structures to fully parse them.
-            - if response_data is a dict and response_model is a dict with specified key and value types:
-                - each key-value pair in the response_data dict is parsed, with the keys being converted to the specified key type, and the values being recursively parsed according to the specified value type.
+            - if response_data is a dict or list, response_data is parsed into an instance of the response_model.
+            - raises error otherwise
```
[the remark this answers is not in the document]
Schmaexxi: it should be both

tests/test_parse_response_model.py
Schmaexxi: thanks! 🧐 will check that again

hari_client/models/models.py
Schmaexxi: That's the version we had before. We had to build this parsing logic ourself, which made the parser unnecessarily complex. This is tested code by a widely used framework.
[Its usage here is exactly like its documented](https://docs.pydantic.dev/latest/concepts/models/#dynamic-model-creation).

Do you think it would help if we explicitly define the root-field?

hari_client/models/models.py
katia-qm: Yeah I agree, but I found out that it was already not breaking a client with new fields, it was the error on the Matthias's side. So probably we don't need this base response.
And your tests currently adds an extra field to the model, but it should add en extra field to the **api response** itself that it not in the model, and here is where it should not break. Do you understand what I mean?

hari_client/client/client.py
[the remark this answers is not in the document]
katia-qm: ah yep, I misunderstood the terminology

tests/test_parse_response_model.py
katia-qm: Pipe is used for python 3.10+, as our min version is 3.11, we could safely use it.
I used a typing.Union earlier because previously we supported 3.9+.

Would be nice to check and replace everything with pipe

hari_client/client/client.py
```diff
@@ -1431,7 +1399,9 @@ def trigger_histograms_update_job(
         Returns:
             models.UpdateHistogramsResponse: updated histograms
         """
-        params = {"compute_for_all_subsets": compute_for_all_subsets}
+        params: dict[str, bool | str] = {
+            "compute_for_all_subsets": compute_for_all_subsets
+        }
```
katia-qm, line 1404: Trigger histograms breaks because it should be a list of UpdateHistograms.
We probably missed that and the parser didn't complain

hari_client/models/models.py
katia-qm: Now I finally get how it works, thank you. Makes sense to me, no more questions

hari_client/models/models.py
Schmaexxi: Great insight about the tests/functions, thanks!
Since we don't have integration tests using a mock http server, yet, I transformed this into a unit test for `_parse_response_model`.

About this not breaking before:
You are absolutely correct that it didn't break but it withholds information from the client/user, since the default setting for extra fields is `ignore` ([see here](https://docs.pydantic.dev/latest/api/config/#pydantic.config.ConfigDict.extra)). This means currently, the client would not break but users wouldn't even know that there wold be extra information returned from an endpoint even when they would inspect a pydantic model as a response from an API endpoint.

Schmaexxi: https://github.com/quality-match/hari-client/pull/11/commits/08df0919b7b155f3932369abfdf95d320cd19695

tests/test_parse_response_model.py
Schmaexxi: we can do this as a follow up, I wouldn't want to mix this into this PR

hari_client/client/client.py
Schmaexxi: good catch, thanks!
https://github.com/quality-match/hari-client/pull/11/commits/57efd756c1ca8bfce8279582b81ba5a3db4efb64

tests/test_parse_response_model.py
Schmaexxi: [508464b](https://github.com/quality-match/hari-client/pull/11/commits/508464b9afaebbff205b3c89d74a5b232bea73b1)

hari_client/models/models.py
katia-qm: Oh I get it now, thanks. However, when I debug the test, I can see it in model_extra, although I can still access it like you did in the test. Do you think users will ever look here? And do you think it's good to show extra fields that are not in the defined model of the client?

hari_client/client/client.py
```diff
@@ -44,88 +37,54 @@ def _parse_response_model(
...
-                return None
+        if response_model is None and response_data is None:
+            return None
+        elif response_model is None or response_data is None:
             raise errors.ParseResponseModelError(
                 response_data=response_data,
                 response_model=response_model,
-                message=f"Expected response_data to be None, but received {response_data=}",
+                message=f"Failed to parse response_data into response_model {response_model}. {response_data=}. Cannot parse None type into a non-None type.",
             )
```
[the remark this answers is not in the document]
Schmaexxi: That's what it says, no?

hari_client/models/models.py
Schmaexxi: that's the expected behavior I want to cover in the test.
if you use the default for https://github.com/quality-match/hari-client/blob/508464b9afaebbff205b3c89d74a5b232bea73b1/hari_client/models/models.py#L12
you would see that the test fails

Schmaexxi: I think it's essential that the client just spits out what the API returns. It's not secret information, anyone using the API would get this information anyway. Withholding it is just ground for confusion imo

hari_client/client/client.py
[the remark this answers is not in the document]
katia-qm: I meant it can also be `Cannot parse none-None type into a None type.`

hari_client/models/models.py
katia-qm: I mean the field is not shown as extra_field in dataset_response, but inside of model_extra structure. Do you think the user will ever look for extra fields there? I'm just wondering whether we need to inform users about extra field returned or not

hari_client/client/client.py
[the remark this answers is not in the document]
Schmaexxi: Ah, of course 👍

katia-qm (approved):
Looks good!

The filters for subset creation are not clear for me though. And the quickstart is not updated, will you do it in another pr?

Schmaexxi:
> Looks good!
> 
> The filters for subset creation are not clear for me though. And the quickstart is not updated, will you do it in another pr?

Yep!

hari_client/client/client.py
```diff
@@ -44,88 +37,54 @@ def _parse_response_model(
...
         # handle pydantic models
+        # order is important here, because the RootModel is a subclass of the BaseModel
         if isinstance(response_model, type) and issubclass(
+            response_model, pydantic.RootModel
+        ):
+            root_field_type = typing.get_origin(
+                response_model.model_fields["root"].annotation
+            )
+            if issubclass(root_field_type, list):
+                return response_model(response_data)
```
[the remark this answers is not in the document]
Dens49: > So far, the API endpoints either respond with a dict or a list

there are mutliple endpoints supported in the client already that expect a plain string

PR #12, opened 2024-09-08
Ml/add mypy

Schmaexxi:
# Background

Still a small project, time to introduce MyPy before effort would be too big to warrant.
Why is MyPy great?
- less bugs
- quicker onboarding
- confidence in refactors

# TODO
- use mypy's strict mode and fix related errors

# Changes
Added mypy to pre-commit hooks


Schmaexxi:
outdated

PR #14, opened 2024-09-23
Reapply pr #11 partially

Dens49:
Review and merge only after PR #13 is merged.

This PR reapplies some changes from the reverted PR#11:
- update create_subset method
- allow extra fields in api models

There's also one additional cleanup:
- commit 2818e2b40b29b180e06a920b29b59a617a91e230 removes a workaround from the response parsing logic which was introduced in v0.1.0, when it wasn't possible to explicitly define `list[MyModel]` as the response type. This has been enabled by PR #1 though. See changelog for more details and overview


alvarotroya (approved):
lgtm 👍🏽

PR #16, opened 2024-09-27
Update hari uploader interface to use bulk_operation_annotatable_id

Dens49:
- Back references aren't required to be unique anymore. Empty and duplicate back_references will log a warning though
- Back references aren't used in the bulk media and media_object upload endpoints anymore. Instead a randomly generated and temporary bulk_operation_annotatable_id is used
- Update quickstart script to fetch processing jobs via trace_id

### DIFF
--- a/hari_client/upload/hari_uploader.py
+++ b/hari_client/upload/hari_uploader.py
@@ -18,19 +18,14 @@ class HARIMediaObject(models.BulkMediaObjectCreate):
     # because it's set internally by the HARIUploader
     bulk_operation_annotatable_id: str = ""
 
-    def get_back_reference(self) -> str:
-        """
-        Returns a back reference with which you can clearly identify this media object after it has been uploaded to HARI.
-
-        Returns:
-            str: The back reference
-
-        Raises:
-            HARIMediaObjectMissingBackReferenceError: If the HARIMediaObject doesn't have a back_reference
-        """
-        if not self.back_reference:
-            raise HARIMediaObjectMissingBackReferenceError(self)
-        return self.back_reference
+    @pydantic.field_validator("media_id", "bulk_operation_annotatable_id")
+    @classmethod
+    def field_must_not_be_set(cls, v: str) -> str:
+        if "" != v:
+            raise ValueError(
+                "The field must not be set on object instantiation. It's used and set by HARIUploader internals."
+            )
+        return v
 
 
 class HARIMedia(models.BulkMediaCreate):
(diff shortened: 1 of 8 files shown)


hari_client/upload/hari_uploader.py
```diff
@@ -8,71 +10,66 @@
...
 
 
-class HARIMediaObject(models.MediaObjectCreate):
+class HARIMediaObject(models.BulkMediaObjectCreate):
     # overwrites the media_id field to not be required,
     # because it has to be set after the media has been uploaded
     media_id: str = ""
+    # overwrites the bulk_operation_annotatable_id field to not be required,
+    # because it's set internally by the HARIUploader
+    bulk_operation_annotatable_id: str = ""
```
FBOBecker, line 19: why not | None = None?

hari_client/upload/hari_uploader.py
```diff
@@ -8,71 +10,66 @@
...
+
+    @pydantic.field_validator("media_id", "bulk_operation_annotatable_id")
+    @classmethod
+    def field_must_not_be_set(cls, v: str) -> str:
+        if "" != v:
```
FBOBecker, line 24: you could also go with `if v` here, no?

hari_client/upload/hari_uploader.py
```diff
@@ -84,40 +81,41 @@ class HARIUploader:
     def __init__(self, client: HARIClient, dataset_id: str) -> None:
         self.client: HARIClient = client
         self.dataset_id: str = dataset_id
-        # key: HARIMedia.back_reference
-        self._medias: dict[str, HARIMedia] = {}
+        self._medias: list[HARIMedia] = []
+        self._media_back_references: set[str] = set()
         self._media_object_back_references: set[str] = set()
+        self._media_object_cnt: int = 0
```
FBOBecker, line 87: no more space left for "ou"? :D

hari_client/upload/hari_uploader.py
```diff
@@ -84,40 +81,41 @@ class HARIUploader:
...
+            HARIMediaUploadError: If an unrecoverable problem with the media upload was detected
         """
         for media in args:
-            # check and remember media by its back_reference
-            media_back_reference = media.get_back_reference()
-            if media_back_reference in self._medias:
-                raise DuplicateHARIMediaBackReferenceError(
-                    back_reference=media_back_reference
+            # check and remember media back_references
+            if media.back_reference in self._media_back_references:
```
FBOBecker, line 101: I like this but I also wonder what to do as a user in this case. I mean it is somewhat to late then, no? They would need to update and fix the uploaded stuff I guess.

But probably still better than raising an error. 🤔

hari_client/upload/hari_uploader.py
```diff
@@ -164,12 +161,18 @@ def upload(
     def _upload_media_batch(
         self, medias_to_upload: list[HARIMedia]
     ) -> tuple[models.BulkResponse, list[models.BulkResponse]]:
+        # set bulk_operation_annotatable_id for all medias_to_upload
```
FBOBecker, line 164: kind of self-explanatory - not sure if comment is needed

hari_client/upload/hari_uploader.py
```diff
@@ -199,20 +202,46 @@ def _upload_media_objects_in_batches(
     def _upload_media_object_batch(
         self, media_objects_to_upload: list[HARIMediaObject]
     ) -> models.BulkResponse:
+        # set bulk_operation_annotatable_id for all media_objects_to_upload
```
FBOBecker, line 205: same

hari_client/upload/hari_uploader.py
```diff
@@ -199,20 +202,46 @@ def _upload_media_objects_in_batches(
...
+            filtered_upload_response = list(
+                filter(
+                    lambda x: x.bulk_operation_annotatable_id
+                    == media.bulk_operation_annotatable_id,
+                    media_upload_bulk_response.results,
+                )
+            )
+            if len(filtered_upload_response) == 0:
+                raise HARIMediaUploadError(
+                    f"Media upload response doesn't match expectation. Couldn't find {media.bulk_operation_annotatable_id=} in the upload response."
```
FBOBecker, line 230: I was today years old when I learned about the '=' syntax for f-strings. Thanks for that!

tests/test_hari_uploader.py
```diff
@@ -147,8 +212,9 @@ def test_hari_uploader_creates_single_batch_correctly(mocker):
     assert len(media_object_calls[0].kwargs["media_objects_to_upload"]) == 10
 
 
-def test_hari_uploader_receives_duplicate_media_back_reference():
+def test_warning_for_hari_uploader_receives_duplicate_media_back_reference(mocker):
     # Arrange
+    log_spy = mocker.spy(hari_uploader.log, "warning")
```
FBOBecker, line 217: again what learned - nice

FBOBecker (dismissed):
🥑

hari_client/upload/hari_uploader.py
Dens49: If I was the user and cared about the back_references I think the easiest thing for me would be to fix the back_references, create a new dataset and ignore or delete the prior one. Since HARI doesn't enforce unique back_references, the warnings are probably our best option. I'll put a small explanation for that in the documentation website

hari_client/upload/hari_uploader.py
FBOBecker: always consider it an abbreviation for a curse word :D

hari_client/upload/hari_uploader.py
FBOBecker: yeah sounds good

FBOBecker (approved):

PR #18, opened 2024-10-01
CRUD methods for attributes

mylius:
Adds CRUD methods for attributes to the client.

### DIFF
--- a/hari_client/client/client.py
+++ b/hari_client/client/client.py
@@ -1758,15 +1758,15 @@ class HARIClient:
             "GET",
             f"/datasets/{dataset_id}/attributes/{attribute_id}",
             params=self._pack(locals(), ignore=["dataset_id", "attribute_id"]),
-            success_response_item_model=models.Attribute,
+            success_response_item_model=models.AttributeResponse,
         )
 
     def update_attribute(
         self,
         dataset_id: str,
         attribute_id: str,
+        annotatable_id: str,
         name: str | None = None,
-        annotatable_id: str | None = None,
         annotatable_type: models.DataBaseObjectType | None = None,
         attribute_group: models.AttributeGroup | None = None,
         value: models.typeT | None = None,
@@ -1822,7 +1822,7 @@ class HARIClient:
             question: The question value
             archived: The archived value
             range: The range value
-            ml_predidictions: The parameters of the posterior Dirichlet distribution
+            ml_predictions: The parameters of the posterior Dirichlet distribution
             ml_probability_distributions: The the Dirichlet distribution values
 
         Returns:
--- a/hari_client/client/client.py
+++ b/hari_client/client/client.py
@@ -1634,6 +1634,76 @@ class HARIClient:
             success_response_item_model=models.BulkResponse,
         )
 
+    def create_attribute(
+        self,
+        id: str,
+        dataset_id: str,
+        name: str,
+        annotatable_id: str,
+        value: models.typeT,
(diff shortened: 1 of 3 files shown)


hari_client/client/client.py
```diff
@@ -1642,3 +1644,153 @@ def create_attributes(
...
+            The attribute with the given attribute_id
+
+        Raises:
+            APIException: If the request fails.
+        """
+        return self._request(
+            "GET",
+            f"/datasets/{dataset_id}/attributes/{attribute_id}",
+            params=self._pack(locals(), ignore=["dataset_id", "attribute_id"]),
+            success_response_item_model=models.Attribute,
```
katia-qm, line 1702: Why does `get_attribute` return `model.Attribute`, but `get_attributeS` returns `list[models.AttributeResponse]`?

katia-qm (changes requested):
Create a separate attribute is missing (not bulk).
What about attribute values/metadata? I'm not sure whether they are needed or not

Comparing all the fields is very overwhelming, so I have to trust you here;)
Did you test all of the methods, to make sure everything works and the parsing doesn't break?

hari_client/client/client.py
mylius: Oversight on my part.

mylius:
> Create a separate attribute is missing (not bulk). What about attribute values/metadata? I'm not sure whether they are needed or not

Having people deal with values and metadata is probably a bit overwhelming, which is why we created the attribute endpoints, which wrap both of those.

> Comparing all the fields is very overwhelming, so I have to trust you here;) Did you test all of the methods, to make sure everything works and the parsing doesn't break?

I did do all CRUD operations at least once for one Attribute, but I didn't populate all fields and try all variations.

hari_client/client/client.py
```diff
@@ -1631,3 +1633,231 @@ def create_attributes(
...
+        median: typing.Any | None = None,
+        variance: typing.Any | None = None,
+        standard_deviation: typing.Any | None = None,
+        range: typing.Any | None = None,
+        average_absolute_deviation: typing.Any | None = None,
+        cumulated_frequency: typing.Any | None = None,
+        frequency: dict[str, int] | None = None,
+        question: str | None = None,
+        archived: bool | None = None,
+        ml_predidictions: dict[str, float] | None = None,
```
katia-qm, line 1792: ml_predi_di_ctions :) also in docstring

hari_client/client/client.py
```diff
@@ -1631,3 +1633,231 @@ def create_attributes(
...
+            params=self._pack(locals(), ignore=["dataset_id", "attribute_id"]),
+            success_response_item_model=models.Attribute,
+        )
+
+    def update_attribute(
+        self,
+        dataset_id: str,
+        attribute_id: str,
+        name: str | None = None,
+        annotatable_id: str | None = None,
```
katia-qm, line 1769: annotatable id seems to be required on the api, could you please double-check?

katia-qm (changes requested):

hari_client/models/models.py
```diff
@@ -921,6 +921,93 @@ def populate_derived_field(cls, values):
         return values
 
 
+class Attribute(BaseModel):
+    id: str = pydantic.Field(title="ID")
+    name: str
```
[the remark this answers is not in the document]
mylius: Not really, but I also don't think it is currently actively used.

katia-qm (approved):
Amazing, thanks a lot!

PR #19, opened 2024-10-01
Km/hari client/apply feedback

katia-qm:
- Update code with 3.11 syntax:
    -remove type in args section of docstring
    - replace typing.Optional with pipe
    - replace typing.Union with pipe
    -
- Improve quickstart example
- Remove linter errors in quickstart


Dens49 (approved):
Thanks for the effort, looks good! Also looks like you caught some str vs uuid inconsistencies between client and backend api :+1: 

nitpick: the cleanups can be mentioned in the changelog for the next release in the `internal` section.

PR #22, opened 2024-10-08
Make attribute group optional defaulting to initial

katia-qm:
Make attribute group optional defaulting to initial when creating attribute(s)


mylius (approved):
Thanks for improving on this!

PR #23, opened 2024-10-09
Fix str-uuid inconsistencies for attribute methods

katia-qm:


mylius (approved):
Thanks for fixing this!

PR #24, opened 2024-10-09
Make media object source optional defaulting to reference

katia-qm:
Make media object source optional defaulting to reference during media object creation


mylius (approved):
Nice improvement, thanks!

PR #26, opened 2024-10-15
Setup of object category subsets in hari uploader

Schmaexxi:
1648824166

# Description
 Users should be able to specify object categories and directly translate them into HARI as subsets.
This PR introduces this functionality

## Changes in this PR
- adds utility-methods to the `HariUploader` used by the `upload`-method
  - _create_object_category_subsets
  - _validate_media_objects_object_category_subsets_consistency
  - _assign_media_objects_to_object_category_subsets
- updates `quickstart.py`
- updates integration tests with object categories
- adds unit tests

# TODO
- [x] Sync with server for existing subsets

### DIFF
--- a/hari_client/models/models.py
+++ b/hari_client/models/models.py
@@ -866,7 +866,7 @@ class MediaObjectCreate(BaseModel):
     scene_id: str | None = None
     realWorldObject_id: str | None = None
     visualisations: list[VisualisationUnion] | None = None
-    subset_ids: set[str] | list[str] | None = None
+    subset_ids: list[str] | None = None
 
     instance_id: str | None = None
     object_category: str | None = None
--- a/hari_client/models/models.py
+++ b/hari_client/models/models.py
@@ -866,7 +866,7 @@ class MediaObjectCreate(BaseModel):
     scene_id: str | None = None
     realWorldObject_id: str | None = None
     visualisations: list[VisualisationUnion] | None = None
-    subset_ids: list[str] | None = None
+    subset_ids: set[str] | list[str] | None = None
 
     instance_id: str | None = None
     object_category: str | None = None
--- a/hari_client/upload/hari_uploader.py
+++ b/hari_client/upload/hari_uploader.py
@@ -126,7 +126,6 @@ class HARIUploader:
     ) -> None:
         self.client: HARIClient = client
         self.dataset_id: str = dataset_id
-        self.object_categories: set[str] = set()
         self.object_categories_to_validate = object_categories_to_validate or set()
         self._medias: list[HARIMedia] = []
         self._media_back_references: set[str] = set()
@@ -220,7 +219,7 @@ class HARIUploader:
                             errors.append(
                                 HARIMediaObjectUnknownObjectCategorySubsetNameError(
                                     f"A subset for the specified object_category_subset_name ({media_object.object_category_subset_name}) wasn't specified."
-                                    f"Only the object_categories that were specified in the HARIUploader constructor are allowed: {self.object_categories}"
+                                    f"Only the object_categories that were specified in the HARIUploader constructor are allowed: {self.object_categories_to_validate}"
                                     f"media_object: {media_object}"
                                 )
                             )
@@ -242,7 +241,7 @@ class HARIUploader:
                         if media_object.object_category is None:
                             raise HARIMediaObjectUnknownObjectCategorySubsetNameError(
                                 f"A subset for the specified object_category_subset_name ({media_object.object_category_subset_name}) wasn't created."
-                                f"Only the object_categories that were specified in the HARIUploader constructor are allowed: {self.object_categories}"
+                                f"Only the object_categories that were specified in the HARIUploader constructor are allowed: {self.object_categories_to_validate}"
                                 f"media_object: {media_object}"
                             )
                         # also add the object_category subset_id to the overall list of subset_ids
@@ -292,7 +291,6 @@ class HARIUploader:
                 f"Found {len(media_object_object_category_subset_errors)} errors with object_category_subset_name consistency."
             )
             raise media_object_object_category_subset_errors
-        # TODO: also validate that each media has the object_category_subset_names of its medias
 
         object_category_subsets = self._get_existing_object_category_subsets()
         # add already existing subsets to the object_category_subsets dict
--- a/hari_client/upload/hari_uploader.py
+++ b/hari_client/upload/hari_uploader.py
@@ -173,9 +173,9 @@ class HARIUploader:
                 self._media_object_cnt += 1
                 self._attribute_cnt += len(media_object.attributes)
 
-    def _create_object_category_subsets(self) -> None:
+    def _create_object_category_subsets(self, object_categories: list[str]) -> None:
         # sort object_categories to ensure consistent subset creation order
-        for object_category in sorted(self.object_categories):
+        for object_category in sorted(object_categories):
             subset_id = self.client.create_subset(
                 dataset_id=self.dataset_id,
                 subset_type=models.SubsetType.MEDIA_OBJECT,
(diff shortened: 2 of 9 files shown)


hari_client/models/models.py
```diff
@@ -866,7 +866,7 @@ class MediaObjectCreate(BaseModel):
     scene_id: str | None = None
     realWorldObject_id: str | None = None
     visualisations: list[VisualisationUnion] | None = None
-    subset_ids: set[str] | list[str] | None = None
+    subset_ids: list[str] | None = None
```
Dens49, line 869: is that also a list in the backend model? For MediaCreate the subset_ids field is still set or list

hari_client/upload/hari_uploader.py
```diff
@@ -156,6 +173,138 @@ def add_media(self, *args: HARIMedia) -> None:
...
+            if obj_cat
+            not in [obj_cat_subset.name for obj_cat_subset in object_category_subsets]
+        ]
+
+        self._create_object_category_subsets(object_categories_without_existing_subsets)
+        log.info(
+            f"All object category subsets of this dataset: {self._object_category_subsets=}"
+        )
+        self._assign_object_category_subsets()
+        # TODO: test this with some already created obj category subsets where others are missing
```
Dens49, line 304: nitpick: I think that's already done?

Schmaexxi: I don't think we have a test case for this yet, do we?

hari_client/models/models.py
Schmaexxi: The type hint in the backend is the same as it was before I made the change, however, [we are now using specific methods of one data structure,](https://github.com/quality-match/hari-client/blob/b9b290400b37f41240b04f1c44d1a5ac262c9083/hari_client/upload/hari_uploader.py#L250) which eliminates the possibility of this being a set

Dens49: Ah I see. The typehint chosen by the backend is a bit unfortunate. Following that approach you should update the typehint of MediaCreate::subset_ids, too. Instead of doing that, could you undo the change to the MediaObjectCreate model? I'd rather to keep it consistent with the backend and look into making this a refactor in backend and client in tandem. Otherwise the interfaces drift (not a huge impact in this example, because set or list makes no difference when you're serializing to json, but still).
In general: A model like MediaObjectCreate that can be used directly or by multiple utilities (not just the hari_uploader) shouldn't be adjusted to fit the implementation details of just one utility.

hari_client/upload/hari_uploader.py
Dens49: ah right, let's do them in a separate PR

hari_client/upload/hari_uploader.py
```diff
@@ -266,6 +268,14 @@ def _get_existing_object_category_subsets(self) -> list[models.DatasetResponse]:
         log.info(f"All existing object_category subsets: {object_category_subsets=}")
         return object_category_subsets
 
+    def get_distinct_subset_names_from_added_media_objects(self) -> set[str]:
```
Dens49, line 271: nitpick: this function could also be passed a list of HARIMedia instead of relying on the state of self._medias

hari_client/upload/hari_uploader.py
```diff
@@ -282,17 +292,23 @@ def _handle_object_categories(self) -> None:
                 f"Found {len(media_object_object_category_subset_errors)} errors with object_category_subset_name consistency."
             )
             raise media_object_object_category_subset_errors
+        # TODO: also validate that each media has the object_category_subset_names of its medias
```
Dens49, line 295: I'm confused by this TODO. There's a typo and you mean "... of its media_objects" right?
But still, how would you check that? You're adding the object category subset ids to medias only later with the call to self._assign_object_category()

hari_client/upload/hari_uploader.py
```diff
@@ -122,11 +122,12 @@ def __init__(
...
         client: HARIClient,
         dataset_id: uuid.UUID,
-        object_categories: set[str] | None = None,
+        object_categories_to_validate: set | None = None,
     ) -> None:
         self.client: HARIClient = client
         self.dataset_id: str = dataset_id
-        self.object_categories: set[str] | None = object_categories
+        self.object_categories: set[str] = set()
+        self.object_categories_to_validate = object_categories_to_validate or set()
```
Dens49, line 130: I don't understand the need for having two fields. Was this intended as a rename? If so, I'd rather to keep calling the function argument only `object_categories`, otherwise the added context from `_to_validate` can create a wrong assumption about this only being used for validation. Makes sense that the function argument is optional, though.
If I saw correctly, the field `self.object_categories` is now only being used in error messages, and is always empty

hari_client/models/models.py
Schmaexxi: Good point, I agree - I'll try to keep the usage of this field agnostic of a more specific type than an iterable as well

hari_client/upload/hari_uploader.py
Schmaexxi: Good catch - it's not needed anymore - I'll remove it. It wasn't meant as a rename but a change in functionality.
It should be optional to pass `object_categories_to_validate`, so users can still create an uploader and upload a dataset without objecet_categories. I find this cleaner and easier to use than requiring an empty list in this case.

We can infer object_categories from the media_objects, anyway. If someone would like to check if their dataset is  consistent, theycan use the validation. I intentionally left it in the quickstart-example, though, so people find the utility without going in-depth

hari_client/models/models.py
katia-qm: I think the set is not even json serialisable, so the request breaks. One point to improve on the hari client.
P.S. uuid is also not json serialisable, we replaced it with a string

Dens49: plan refactor:
- backend and client models for subset_ids to be list
- client should have custom serialization options for json to allow for UUID to be serialized

hari_client/upload/hari_uploader.py
```diff
@@ -157,6 +186,156 @@ def add_media(self, *args: HARIMedia) -> None:
...
+            media_object_object_category_subset_name_errors,
+        ) = self._get_and_validate_media_objects_object_category_subset_names()
+        if media_object_object_category_subset_name_errors:
+            log.error(
+                f"Found {len(media_object_object_category_subset_name_errors)} errors with object_category_subset_name consistency."
+            )
+            raise ExceptionGroup(
+                f"Found {len(media_object_object_category_subset_name_errors)} errors with object_category_subset_name consistency.",
+                media_object_object_category_subset_name_errors,
+            )
```
Schmaexxi, line 313: Users are missing out on what the details of the error are and how they can fix it. Would be nice if they either get a log or some other way of knowing which categories clash and what the associated objects are

tests/upload/fixtures.py
```diff
@@ -0,0 +1,363 @@
...
+        uploader,
+        "_set_bulk_operation_annotatable_id",
+        side_effect=id_setter_mock,
+    )
+    id_setter_spy = mocker.spy(uploader, "_set_bulk_operation_annotatable_id")
+
+    return uploader, id_setter_spy
+
+
+def create_configurable_mock_uploader_successfull_single_batch(
```
Schmaexxi, line 213: NIT
```suggestion
def create_configurable_mock_uploader_successful_single_batch(
```

tests/upload/fixtures.py
```diff
@@ -0,0 +1,363 @@
...
+    object_category_subset_ids: dict[str, uuid.UUID] | None = None,
+) -> tuple[
+    hari_uploader.HARIUploader,
+    HARIClient,
+    typing.Any,
+    typing.Any,
+    typing.Any,
+    typing.Any,
+]:
+    """Creates a configurable mock uploader for a successfull upload of a single batch of medias, media objects and attributes.
```
Schmaexxi, line 230: NIT
```suggestion
    """Creates a configurable mock uploader for a successful upload of a single batch of medias, media objects and attributes.
```

tests/upload/fixtures.py
Schmaexxi, line 213: NIT: I'd prefer fixtures to be decorated as such, so we can use them as arguments in test-functions.
For this approach, a factory function would work https://docs.pytest.org/en/6.2.x/fixture.html#factories-as-fixtures

tests/test_hari_uploader.py
```diff
@@ -647,6 +688,200 @@ def id_setter_mock(item: hari_uploader.HARIMedia | hari_uploader.HARIMediaObject
...
+    media_object_3 = hari_uploader.HARIMediaObject(
+        source=models.DataSource.REFERENCE, back_reference="img_1_obj_3"
+    )
+
+    media_1 = hari_uploader.HARIMedia(
+        name="my image 1", media_type=models.MediaType.IMAGE, back_reference="img_1"
+    )
+
+    # Act + Assert
+    # "waste" the first mocked call to get_subsets_for_dataset which returns an empty list
```
Schmaexxi, line 875: NIT: instead of having this logic in the test, it would be nicer have this in the fixture setting this up. Could be another parameter. Like this, it makes the fixture, which is supposed to be a more-or-less generic factory quite restrictive by being bound to the return value of the `get_subsets_for_dataset`-endpoint

Schmaexxi (commented):
Cannot request changes because I opened the PR. I got one concern about usability and error logging. The rest looks good and is nitpicking, which would be nice to have, imo

hari_client/upload/hari_uploader.py
Schmaexxi: my bad, I was only looking at the test output 👍

tests/upload/fixtures.py
Dens49: ah thanks :+1:
was actually looking through the documentation for exactly this, but couldn't find it myself

hari_client/upload/hari_uploader.py
Dens49: actually not needed anymore, so I've sent it to the shadow realm :black_circle:

tests/test_hari_uploader.py
Dens49: adds a bit more boilerplate to the usage of the uploader mock, but I'm fine with that

Schmaexxi:
Thanks @Dens49, looks good!

PR #27, opened 2024-10-17
fix shared media object/attribute correct media/annotatable id update

katia-qm:
When a media object for media or an attribute for media/media object is shared, this shared attribute/object for different media/media object is updated all the times that any of the attributes is updated, so at the end all attributes that were shared receive the last media/annotatable id and it created duplicated attribute (with the same attribute id and annotatable id)

The pr update fixes the issue


mylius (approved):
Thanks for discovering and fixing this!

PR #29, opened 2024-10-21
Restrict object_category field to UUID

mylius:
The hari API no longer allows the object_category to be arbitrary strings, but it rather has to be a UUID.

This PR adjusts the client to match this behavior and adds a custom JSONEncoder, that is able to serialize UUIDs.

### DIFF
--- a/hari_client/client/client.py
+++ b/hari_client/client/client.py
@@ -1,4 +1,5 @@
 import datetime
+import json
 import pathlib
 import types
 import typing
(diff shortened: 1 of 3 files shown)


hari_client/client/client.py
```diff
@@ -1177,7 +1188,7 @@ def update_media_object(
             media_id: Media Id
             instance_id: Instance Id
             source: DataSource
-            object_category: Object Category
+            object_category: Object Categories subset id
```
katia-qm, line 1191: maybe
```suggestion
            object_category: Object category's subset id
```
?

katia-qm (approved):
Thanks!

Dens49:
awesome! also thanks for the JSONEncoder, which already implements one of the refactoring ideas from the object_category PR https://github.com/quality-match/hari-client/pull/26#discussion_r1806396627 :+1: Feel free to merge, I'll resolve all conflicts with the object_category PR afterwards

PR #30, opened 2024-10-22
Clean up of uuid vs str inconsistencies in current client api models

Dens49:
1648824166

The inconsistencies weren't really a technical problem, because we now have the custom UUID serializer from #29, but there were pydantic warnings about it when running the quickstart script. They should be gone now with this update.


Schmaexxi (approved):
Looks good!

PR #31, opened 2024-10-22
Fix update attribute request

Schmaexxi:
# Background
Request errored out because server expected different model.

# Changes
Fix update_attribute-method


katia-qm (approved):
Thanks a lot for finding it!

PR #32, opened 2024-10-23
fix `HARIClient.create_attribute` method

Dens49:
the question field gets special treatment in the backend, therefore it must be ignored by the client if it's None


katia-qm (approved):
Elegant!

PR #35, opened 2024-10-24
Make use of empty subset creation endpoint

mylius:
Hari has multiple endpoints for creating subsets.
The one we were using so far contains more logic and takes longer to execute. The extra logic is already covered in metadata_rebuild, which is necessary to be called, anways.


Schmaexxi (approved):
Great, thanks! Could you please update the changelog, too?

PR #38, opened 2024-11-07
Serialize datetime objects to iso 8601 strings

Dens49:
1614747154

Without this fix the serialization of datetime objects fails, because it's not handled automatically.


katia-qm (approved):
Makes sense to me:)

PR #42, opened 2024-11-15
Added support for mixed file extensions in `create_medias`

Dens49:
1711901915

The backend endpoint `/datasets/{dataset_id}/medias/uploadUrl` for the client method `get_presigned_media_upload_url` accepts only a single file_extension. So far the hari-client didn't allow mixed file extensions. In order to support that in `create_medias` and therefore the hari_upload utility, the file paths are now grouped by file extension and there's one call to `get_presigned_media_upload_url` for each file extension group.


Emprime:
Test

tests/test_client.py
```diff
@@ -19,9 +21,8 @@ def test_create_medias_with_missing_file_paths(test_client):
         client.create_medias(dataset_id="1234", medias=[media_create])
 
 
-def test_create_medias_with_different_file_extensions(test_client):
+def test_create_medias_with_different_file_extensions_works(test_client, mocker):
```
Emprime, line 24: Where does this object come from? I can't seem to find it in the file.
It's most likely a mocking tool for testing. Could you give me the details to the used mocking tool?

tests/test_client.py
```diff
@@ -34,10 +35,55 @@ def test_create_medias_with_different_file_extensions(test_client):
...
+    request_spy = mocker.spy(test_client, "_request")
+
+    # Act
+    test_client.create_medias(
+        dataset_id="1234", medias=[media_create_1, media_create_2]
+    )
+
+    # Assert
+    # called twice, because every group of files with the same extension is uploaded in a separate batch
+    assert get_presigned_media_upload_url_spy.call_count == 2
```
Emprime, line 59: This I'm not familiar with the tooling, I can not evaluate if this test is correct but looks good.

Emprime (approved):
Looks good to me, the changes are clear and all necessary

tests/test_client.py
Dens49: test_client and mocker are both injected as pytest fixtures. `test_client` is our self defined fixture and `mocker` is defined by `pytest-mock` https://pytest-mock.readthedocs.io/en/latest/usage.html 
You can find our self defined fixtures in `conftest.py` - they're either implemented directly in that file or imported there. That's how pytest knows about fixtures and is able to auto-inject them into your test functions. https://docs.pytest.org/en/stable/reference/fixtures.html#conftest-py-sharing-fixtures-across-multiple-files

tests/test_client.py
Dens49: test spies are useful when you want to check the behavior of a piece of code. With this spy I'm checking whether the endpoint method is called as often as I would expect, based on how I want the code under test supposed to behave.

katia-qm (approved):
Good job! Left tiny change requests

PR #43, opened 2024-11-22
Added Scripts

Emprime:
These are the first two scripts of a list of potential common use case scripts. 
The idea is that you can use the scripts in three major ways:
1. Execute the script directly with python. Please remember to add the `.env` to your working directory.
2. Use the defined functions in larger scripts
3. Take the code as inspiration for your personalized scripts.

If a script needs an extended client capability, these can be added during the same PR. I added, for example, the ability to handle AttributeMetadata and Get Visualisation Configs.


scripts/create_subsets_from_attribute.py
```diff
@@ -0,0 +1,131 @@
...
+    # Construct a query to find the attribute with the specified ID.
+    query: dict = {"attribute": "id", "query_operator": "==", "value": attribute_id}
+
+    # Execute the query to retrieve the attributes.
+    attributes = hari.get_attributes(dataset_id, query=json.dumps(query))
+    print(f"Found {len(attributes)} attributes with ID {attribute_id}")
+
+    # 3. Extract needed subsets.
+    # Collect all unique labels from the attribute frequencies.
+    labels: List[str] = list(
```
Dens49, line 42: please use typehints with built-in types like `list[xyz]` instead of imported types from `typing` (this is possible since python 3.9 https://docs.python.org/3/whatsnew/3.9.html#type-hinting-generics-in-standard-collections)

scripts/create_subsets_from_attribute.py
```diff
@@ -0,0 +1,131 @@
...
+) -> None:
+    """
+    Function to create subsets in a HARI dataset based on a given attribute.
+    Disclaimer: This function only works for MediaObjects subsets. It uses the first visualization config by default.
+
+    Args:
+        hari (HARIClient): API Client
+        dataset_id (str): The ID of the dataset to work on.
+        attribute_id (str): The ID of the attribute to query.
+        prefix (str): A prefix to add to subset names for better sorting.
```
Dens49, line 23: since function args are already typehinted, you shouldn't specify the types in the docstring again (google style guide)

scripts/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+Please be aware that this folder is currently under construction and may include outdated code.
+We are working on including more use cases in the future.
+Below you can find a description of the different available scripts.
+
+## Available scripts
+
+### Subsets
+
+- `create_subsets_from_attribute.py` Creates subsets in a HARI dataset based on all aggregated values of an attribute.
```
Dens49, line 18: an example command to run it would be nice

Dens49 (commented):
just a few more observations

hari_client/client/client.py
```diff
@@ -1944,3 +1944,72 @@ def delete_attribute(
...
+            The deleted attribute
+
+        Raises:
+            APIException: If the request fails.
+        """
+        return self._request(
+            "DELETE",
+            f"/datasets/{dataset_id}/attributeMetadata/{attribute_id}",
+            params=self._pack(locals(), ignore=["dataset_id", "attribute_id"]),
+            success_response_item_model=(str | None),
```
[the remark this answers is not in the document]
Emprime: It states here, return value is string for 200 https://bbb.qm-annotations.com/docs#/attributes/archive_attribute_datasets__dataset_id__attributeMetadata__attribute_id__delete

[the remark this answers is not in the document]
Dens49: ah, I see that the openapi docs define the schema as `any` for this case. The `"string"` you see there is just an example value.
So the proper typehint would be `typing.Any`. The response parser doesn't support that, though... I'll add that to the list of improvements.
Could you write it as `str | int | float | list | dict | None` please so that it's as close as possible to any?

Dens49 (approved):
we agreed that I finish up the PR with the final small change request. Also cleaned up the changelog, which was necessary after updating this branch to the latest main

PR #47, opened 2024-11-27
Add repeats, possible values and cant solve ratio to all attribute models

katia-qm:
Add repeats and possible values, as well as cant solve ration, to all attribute models


mylius (approved):
Would you do me a favor and also add the cant_solves ratio?

mylius (commented):
Thanks, looks good to me.

PR #49, opened 2024-12-04
Fix query arg serialization

Dens49:
1721653103

The `query` argument wasn't serialized properly for multiple methods.
This is now fixed. Support for existing workarounds, where the query argument was passed as an already stringified single QueryParameter oder LogicParameter object or as a list of those is kept, but will be removed in the future. If you've already used this workaround, please update your scripts.

### DIFF
--- a/hari_client/client/client.py
+++ b/hari_client/client/client.py
@@ -2073,6 +2073,11 @@ class HARIClient:
     def get_visualisation_configs(
         self,
         dataset_id: uuid.UUID,
+        archived: bool | None = False,
+        query: models.QueryList | None = None,
+        sort: list[models.SortingParameter] | None = None,
+        limit: int | None = None,
+        skip: int | None = None,
     ) -> list[models.VisualisationConfiguration]:
         """
         Retrieve the visualization configurations for a given dataset.
--- a/hari_client/client/client.py
+++ b/hari_client/client/client.py
@@ -141,9 +141,10 @@ def handle_union_parsing(item, union_type):
 def _serialize_query_list_for_request(
     query: models.QueryList | list[str] | str | None,
 ) -> list[str] | None:
-    """Serialize a QueryList to a list of strings that can be used as a list query parameter in the HARIClient::_request method.
-    In the future only a QueryList should be supported, but due to existing workarounds, this method also supports those workarounds.
-    You can pass a single already serialized QueryParameter and LogicParameter object (serialized with json.dumps), or a list of them.
+    """Serializes a variable of type QueryList to a list[str]. This list can be passed to the HARIClient::_request method as a query parameter and it will handle
+    formatting it as a query parameter array.
+    Note that in the future only QueryList will be supported. For now other types are supported due to existing workarounds.
+    The workarounds are: passing a single already serialized QueryParameter/LogicParameter object (serialized with json.dumps), or a list of them.
 
     Args:
         query: The QueryList to serialize
--- a/hari_client/client/client.py
+++ b/hari_client/client/client.py
@@ -138,45 +138,52 @@ def handle_union_parsing(item, union_type):
     )
 
 
-def _serialize_query_list_for_request(
-    query: models.QueryList | list[str] | str | None,
-) -> list[str] | None:
-    """Serializes a variable of type QueryList to a list[str]. This list can be passed to the HARIClient::_request method as a query parameter and it will handle
-    formatting it as a query parameter array.
-    Note that in the future only QueryList will be supported. For now other types are supported due to existing workarounds.
-    The workarounds are: passing a single already serialized QueryParameter/LogicParameter object (serialized with json.dumps), or a list of them.
+def _prepare_request_query_params(
+    params: dict[str, typing.Any]
+) -> dict[str, typing.Any]:
+    """Prepares query parameters for the request modules `request` method.
+    Handled cases:
+      - parameter value is a list of pydantic models.
+        - Serializes a query param value of type list[pydantic.BaseModel] to a list[str]. Lists are formatted by the `request` method as `?my_list=value_1&my_list=value_2&my_list=value_3...`,
+        but it doesn't automatically serialize a list of pydantic models, so we have to handle this here.
+        This method contains a workaround for the param "query". It's expected type is QueryList.
+            - The workarounds are: passing a single already serialized QueryParameter/LogicParameter object (serialized with json.dumps), or a list of them.
+            - Note that in the future only QueryList will be supported for query. For now other types are supported due to existing workarounds.
 
     Args:
-        query: The QueryList to serialize
+        params: The query parameters that should be added to the request.
+
+    Returns:
+        The query parameters dictionary with properly serialized values.
     """
-    if query is None:
-        return None
+    params_copy = {}
 
-    if isinstance(query, list):
-        serialized_query = []
-        for x in query:
-            if isinstance(x, models.QueryParameter) or isinstance(
-                x, models.LogicParameter
-            ):
-                serialized_query.append(json.dumps(x.model_dump()))
-            elif isinstance(x, str):
-                serialized_query.append(x)
-                msg = (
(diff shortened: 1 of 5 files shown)


tests/test_client.py
```diff
@@ -250,3 +252,109 @@ def test_trigger_metadata_rebuild_validation_for_dataset_ids_list(test_client):
         client.trigger_metadata_rebuild_job(
             dataset_ids=[1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11],
         )
+
+
+@pytest.mark.parametrize(
+    "query, expected",
```
katia-qm, line 258: nitpick: could you also add query=None test case please?

hari_client/client/client.py
```diff
@@ -832,6 +874,9 @@ def get_medias(
         Raises:
             APIException: If the request fails.
         """
+
+        query = _serialize_query_list_for_request(query)
+
```
katia-qm, line 879: do you think there is a way to move this somewhere in the _request and check if query is in params?

hari_client/client/client.py
```diff
@@ -136,6 +138,46 @@ def handle_union_parsing(item, union_type):
...
+            ):
+                serialized_query.append(json.dumps(x.model_dump()))
+            elif isinstance(x, str):
+                serialized_query.append(x)
+                msg = (
+                    "Argument's 'query' content was detected to be a string, but should be QueryParameter or LogicParameter."
+                    + " Support for this behavior will be removed in a future release."
+                )
+                warnings.warn(msg)
+                log.warning(msg)
```
katia-qm, line 168: why two warnings, what is the difference?

hari_client/client/client.py
```diff
@@ -136,6 +138,46 @@ def handle_union_parsing(item, union_type):
...
+        return None
+
+    if isinstance(query, list):
+        serialized_query = []
+        for x in query:
+            if isinstance(x, models.QueryParameter) or isinstance(
+                x, models.LogicParameter
+            ):
+                serialized_query.append(json.dumps(x.model_dump()))
+            elif isinstance(x, str):
```
katia-qm, line 161: the user can pass here any string, if so, we get an error on the api. We should just let it be, right?

katia-qm: btw we should probably improve the error message there

hari_client/client/client.py
```diff
@@ -136,6 +138,46 @@ def handle_union_parsing(item, union_type):
...
+def _serialize_query_list_for_request(
+    query: models.QueryList | list[str] | str | None,
+) -> list[str] | None:
+    """Serialize a QueryList to a list of strings that can be used as a list query parameter in the HARIClient::_request method.
+    In the future only a QueryList should be supported, but due to existing workarounds, this method also supports those workarounds.
+    You can pass a single already serialized QueryParameter and LogicParameter object (serialized with json.dumps), or a list of them.
+
+    Args:
+        query: The QueryList to serialize
+    """
```
katia-qm, line 150: maybe
```suggestion
    """Serialize query list to a list of strings that can be used as a QueryList parameter in the HARIClient::_request method. 
    You can pass QueryList, as well as a single already serialized QueryParameter and LogicParameter object (serialized with json.dumps), or a list of them.
    Note that in the future only QueryList will be supported, for now other types are supported due to existing workarounds.

    Args:
        query: The QueryList to serialize
    """
```

hari_client/client/client.py
```diff
@@ -136,6 +138,46 @@ def handle_union_parsing(item, union_type):
...
+                warnings.warn(msg)
+                log.warning(msg)
+        query = serialized_query
+    elif isinstance(query, str):
+        msg = (
+            "Argument 'query' was passed as a string, but should be passed as a QueryList (list of QueryParameter or LogicParameter objects)."
+            + " Support for this behavior will be removed in a future release."
+        )
+        warnings.warn(msg)
+        log.warning(msg)
```
katia-qm, line 176: maybe we should add an else there? Otherwise we get 500 error with empty message for unsupported types

katia-qm (changes requested):
Good job! I got small remarks and questions as always:)

I noticed that get_vis_config endpoint doesn't have all the arguments that the corresponding api method has, is there any reason for it? If not, can you add the missing arguments? (query arg is among the missed ones)

hari_client/client/client.py
Dens49: thought about that, too, but I'd rather have this responsibility explicitly in the method that knows that it has such a query param. `_request` shouldn't be concerned with such specifics imo.

hari_client/client/client.py
Dens49: the first is a native python warning and the other is just logging the message with the "warning" log level. Depending on the configured log level, the warning might not be logged. I couldn't a definitive answer on whether one should be preferred over the other. Using both doesn't hurt and for example has the effect that in successful pytest runs you still see the warnings, even though logs are not printed for successful runs.

hari_client/client/client.py
Dens49: I'd leave handling unsupported types to the requests module, as it was doing before

hari_client/client/client.py
Dens49: > the user can pass here any string, if so, we get an error on the api. We should just let it be, right?

agreed, the backend will complain if it's not a proper QueryParameter or LogicParameter. Since it's not intended behavior anyway, I wouldn't do extra validation here.

> btw we should probably improve the error message there

which error message where?

hari_client/client/client.py
Dens49: The type naming leads to confusion here. I'll reiterate in the docstring.
The type of the `query` arg is `QueryList` which is a `list` of `QueryParameter` objects (and or LogicParameter). The method serializes a variable of type `QueryList` to a list of strings which the `HARIClient::_request` method can translate into `HTTP query parameters`, which the backend can parse as a `QueryList` again.

hari_client/client/client.py
katia-qm: Yeah, makes sense. Just hope it will not get forgotten for new endpoints

hari_client/client/client.py
katia-qm: hmm and none of them is printed 100% of time?

hari_client/client/client.py
Dens49: > Just hope it will not get forgotten for new endpoints

ouch, that's a really good point. Let me have another look...

hari_client/client/client.py
katia-qm: Isn't using warnings sufficient for all the cases then? I don't really mind having both, but may be confusing for the users that get 2 same warnings (as for me)

Dens49: as long as the user doesn't redirect stderr, the warnings will surface. Only kept warnings.warn now

hari_client/client/client.py
katia-qm: maybe you could also add it to the error improvement ticket you created?

katia-qm (approved):
Good job!

hari_client/client/client.py
```diff
@@ -138,45 +138,52 @@ def handle_union_parsing(item, union_type):
...
-    query: models.QueryList | list[str] | str | None,
-) -> list[str] | None:
-    """Serializes a variable of type QueryList to a list[str]. This list can be passed to the HARIClient::_request method as a query parameter and it will handle
-    formatting it as a query parameter array.
-    Note that in the future only QueryList will be supported. For now other types are supported due to existing workarounds.
-    The workarounds are: passing a single already serialized QueryParameter/LogicParameter object (serialized with json.dumps), or a list of them.
+def _prepare_request_query_params(
+    params: dict[str, typing.Any]
+) -> dict[str, typing.Any]:
+    """Prepares query parameters for the request modules `request` method.
```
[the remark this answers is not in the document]
Dens49: oh there's an apostrophe missing. `... for the request module's "request" method.`

PR #50, opened 2024-12-04
Configurable batch size and split progressbar

Dens49:
With this update, the batch sizes for the bulk uploads are made configurable.
The defaults are:
- media upload: 30
- media object upload: 500
- attribute upload: 500

Using this is optional. The .env_example shows how to set this with the .env file.

The progressbars are separated so that the actual upload progress of media objects and attributes is made visible, too.


tests/conftest.py
```diff
@@ -13,5 +14,13 @@ def test_client():
...
         hari_client_id="client_id",
         hari_auth_url="auth_url",
+        # tests related to batching were written before batch size
+        # was configurable, so use the old default 500 for all tests
+        # using the test_client fixture
+        hari_uploader=HARIUploaderConfig(
+            media_upload_batch_size=500,
+            media_object_upload_batch_size=500,
+            attribute_upload_batch_size=500,
+        ),
```
katia-qm, line 24: I think it's a good solution for existing tests, but I'd additionally write more tests with different sizes for each case to be sure we are not using still old bulk upload size value for everything

docs/example_code/.env_example
```diff
@@ -1,2 +1,7 @@
 HARI_USERNAME="john.doe@example.com"
 HARI_PASSWORD="MY_SECRET_PASSWORD"
+
+# Optionals
+# HARI_UPLOADER__MEDIA_UPLOAD_BATCH_SIZE=30
+# HARI_UPLOADER__MEDIA_OBJECT_UPLOAD_BATCH_SIZE=500
+# HARI_UPLOADER__ATTRIBUTE_UPLOAD_BATCH_SIZE=500
```
katia-qm, line 7: I think we should mention it in the readme, as we say that there are only hari credentials there

CHANGELOG.md
```diff
@@ -10,6 +10,12 @@
   - get_attribute_metadata
   - delete_attribute_metadata
   - get_visualisation_configs
+- batch sizes for bulk uploads are configurable and the default for media upload was reduced to 30 [PR#50](https://github.com/quality-match/hari-client/pull/50)
+  - see the `.env_example` for how to set the batch sizes with your .env file.
+  - defaults:
+    - media upload: 30 (was 500 previously)
+    - media object upload: 500 (as before)
+    - attribute upload: 500 (as before)
```
katia-qm, line 18: progress bar entry here?

katia-qm, line 18: progress bar change mention here?

CHANGELOG.md
```diff
@@ -10,6 +10,12 @@
   - get_attribute_metadata
   - delete_attribute_metadata
   - get_visualisation_configs
+- batch sizes for bulk uploads are configurable and the default for media upload was reduced to 30 [PR#50](https://github.com/quality-match/hari-client/pull/50)
```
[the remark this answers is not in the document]
Dens49: yes there have been timeouts for slow connections. You'll also get quicker progress feedback for smaller batch sizes - I'd like to also change the progress indication so that after every single media file upload we can increase the progress so that we don't have to wait for the full batch, but not in this iteration, because that needs a bit of a structural change

hari_client/config/config.py
```diff
@@ -25,11 +32,13 @@ class Config(pydantic_settings.BaseSettings):
     """
 
     model_config = pydantic_settings.SettingsConfigDict(
-        env_file=".env", env_file_encoding="utf-8"
+        env_file=".env", env_file_encoding="utf-8", env_nested_delimiter="__"
```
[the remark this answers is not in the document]
Dens49: this defines how pydantic parses the .env file for nested models

tests/conftest.py
Dens49: right there should also be a small test making sure the configured batch size is used

docs/example_code/.env_example
Dens49: the readme doesn't say there are only hari credentials there. It just mentions to fill them in there

hari_client/config/config.py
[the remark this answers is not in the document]
katia-qm: Why not to use the default `_`?

docs/example_code/.env_example
katia-qm: You're right, but at least we should let the users know somewhere that this possibility is there, otherwise it's only inside env example and without any explanation

hari_client/config/config.py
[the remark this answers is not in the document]
Dens49: because we're already using snake_cased env variable names. So the single underscore would technically work (never tested that though), but as a human you'd have a hard time to see the nesting imo

tests/conftest.py
Dens49: I somehow thought there were way more tests for the batching, but it was actually just one. updated that test to use custom configured batch sizes and removed setting the all batch sizes to 500 in the test_client fixture

README.md
```diff
@@ -21,6 +21,30 @@ You can find the referenced images here: `docs/example_code/images/`.
...
+### HARI Uploader configuration
+
+#### Upload batch sizes
+
+You can configure the bath size that's used to upload medias, media objects and attributes.
+If you're experiencing timeout problems, we recommend reducing the media upload batch size.
+
+- `HARI_UPLOADER\_\_MEDIA_UPLOAD_BATCH_SIZE`
+- `HARI_UPLOADER\_\_MEDIA_OBJECT_UPLOAD_BATCH_SIZE`
+- `HARI_UPLOADER\_\_ATTRIBUTE_UPLOAD_BATCH_SIZE`
```
katia-qm, line 46: Are the slashes necessary? they still appear as slashes in readme

katia-qm (approved):
Good job, thanks!

PR #53, opened 2024-12-10
do file upload for create_medias with requests.Session

Dens49:
1704966719

The many PUT requests for the file uploads can reuse an already established TCP connection, which is done by using `requests.Session`. This reduces the overhead of the setup and tear down of a connection for every single file upload.
The other benefit is that the built-in retry mechanism can be applied easily. There was no retry mechanism before, but it's best practice to have one in order to stabilize the upload process


hari_client/client/client.py
```diff
@@ -371,6 +376,32 @@ def _upload_media_files_with_presigned_urls(
...
+        # set up the session with retry mechanism
+        session = requests.Session()
+        # due to the SSLEOFError obscuring the underlying error response from the cloud provider, we don't know
+        # which status code to retry on. Therefore we retry on every 5xx codes, as well as the
+        # two default 4xx codes.
+        retries = adapters.Retry(
+            total=5,
+            backoff_factor=0.1,
+            status_forcelist=[
+                413,
```
mylius, line 388: I don't think the payload will shrink over time.

But I guess that's what the comment about obfuscation is about?!

hari_client/client/client.py
```diff
@@ -371,6 +376,32 @@ def _upload_media_files_with_presigned_urls(
...
+                429,
+                503,
+                500,
+                501,
```
mylius, line 392: Feels unlikely a not implemented method will change the response on a retry.

But I guess that's what the comment about obfuscation is about?!

hari_client/client/client.py
```diff
@@ -371,6 +376,32 @@ def _upload_media_files_with_presigned_urls(
...
+                502,
```
mylius, line 393: nit:
```suggestion
                500,
                501,
                502,
                503,
```

hari_client/client/client.py
```diff
@@ -371,6 +376,32 @@ def _upload_media_files_with_presigned_urls(
                 files_by_file_extension[file_extension] = []
             files_by_file_extension[file_extension].append(file_path)
 
+        # set up the session with retry mechanism
+        session = requests.Session()
+        # due to the SSLEOFError obscuring the underlying error response from the cloud provider, we don't know
+        # which status code to retry on. Therefore we retry on every 5xx codes, as well as the
+        # two default 4xx codes.
+        retries = adapters.Retry(
```
mylius, line 384: I wonder whether it makes sense to generally establish retries, also when talking to our server.

hari_client/client/client.py
```diff
@@ -310,9 +311,13 @@ def _pack(locals_, not_none: list[str] = None, ignore: list[str] = None):
 
         return {k: locals_[k] for k in local_vars if k not in ["self", "kwargs"]}
 
-    def _upload_file(self, file_path: str, upload_url: str) -> None:
+    def _upload_file(
+        self, file_path: str, upload_url: str, session: requests.Session = None
+    ) -> None:
+        if session is None:
+            session = requests.Session()
```
mylius, line 318: Just for my understanding: Are sessions mutable, and that why we can't initialize it as the default?

mylius (approved):
Just minor comments, questions and food for thought. Looks fine to me.

hari_client/client/client.py
Dens49: about the SSLEOFError: there's a giant rabbit hole of issues around the SSLEOFError that Benni and I saw (https://github.com/urllib3/urllib3/issues/3100). Since it's not happening consistently for us, it's very likely not an issue with certificates, we also don't control the cloud provider certificates.
It's likely that this list of status codes might not even be necessary. Probably the server was never reached by the client's request and that's why there's an unexpected EOF error in the ssl handshake :thinking: 
If you're not setting a custom status_forcelist, the default status codes on which retries are triggered are 413, 429, 503 (https://urllib3.readthedocs.io/en/stable/reference/urllib3.util.html#urllib3.util.Retry.RETRY_AFTER_STATUS_CODES)

hari_client/client/client.py
Dens49: yeah they're stateful. Every session object has its own connection pool and can be configured with different settings for different hosts. How I've used the session object is definitely not optimal, especially if you look at the session object that's used for the requests to HARI. You could make it all work with just that one session object, but I didn't want to risk accidentally configuring it to do retries for requests to HARI. Using a separate shorter lived session object seemed less prone to side-effects

hari_client/client/client.py
Dens49: For some endpoints it definitely makes sense. I think we'd have to take a deeper look and also think about request ids, to prevent multi submissions (for example if you have a retry mechanism for an endpoint that takes long to respond and the client is retrying due to a local client timeout or proxy 504)

PR #54, opened 2024-12-12
Fix mixed media file extensions upload

Dens49:
1739954130

Fixes a bug introduced in #42 (not included in any release).
Mixed file extensions led to mixing up the mapping between uploaded image files and medias.

### DIFF
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -37,11 +37,22 @@ def test_create_medias_with_different_file_extensions_works(test_client, mocker)
         media_type=models.MediaType.IMAGE,
         file_path="./my_test_media_2.png",
     )
+    media_create_3 = models.MediaCreate(
+        name="my test media 3",
+        back_reference="my test media 3 backref",
+        media_type=models.MediaType.IMAGE,
+        file_path="./my_test_media_3.jpg",
+    )
     mocker.patch.object(test_client, "_upload_file")
     mocker.patch.object(
         test_client,
         "get_presigned_media_upload_url",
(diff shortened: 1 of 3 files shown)


tests/test_client.py
```diff
@@ -180,28 +207,65 @@ def test_upload_media_files_with_presigned_urls_with_multiple_file_extensions(
...
     # for every call
+    file_paths = {
+        0: "./my_test_media_1.jpg",
+        1: "./my_test_media_2.png",
+        2: "./my_test_media_3.jpg",
+        3: "./my_test_media_4.png",
+    }
+
+    def get_presigned_media_upload_url_side_effect_function(
+        dataset_id, file_extension, batch_size
```
katia-qm, line 218: nitpick: unused dataset id

tests/test_client.py
```diff
@@ -180,28 +207,65 @@ def test_upload_media_files_with_presigned_urls_with_multiple_file_extensions(
...
+    ):
+        response = []
+        for i in range(batch_size):
+            # creates idx 1, 2 in the first call, and 3, 4 in the second call of the method
+            idx = i + get_presigned_media_upload_url_spy.call_count
+            if get_presigned_media_upload_url_spy.call_count > 1:
+                idx += 1
```
katia-qm, line 225: I don't really get how it works, but I trust you:)

katia-qm (approved):
Great job and thanks for the thorough tests!

tests/test_client.py
Dens49: step through it in the debugger, then it should be self evident, but it's really not worth the time

PR #55, opened 2024-12-12
add validations for initial attributes

Dens49:
the validation is run automatically when using the hari_uploader's upload method.
Per default the validation is also run in the client's create_attributes method, but can be disabled with an arg, which the hari_uploader uses to prevent validation from running again for every batch.
Also took the opportunity to set the annotatable_type of attributes in the hari_uploader before the upload happens. Knowing the annotatable type is important for the validation.

### DIFF
--- a/hari_client/client/errors.py
+++ b/hari_client/client/errors.py
@@ -102,7 +102,7 @@ class AttributeValidationInconsistentValueTypeError(Exception):
     ):
         message = (
             f"Found multiple value types {found_value_types} for attribute {attribute_name} with {annotatable_type=}."
-            + " Make sure every attribute with the same name and attribute type has the same value type."
+            + " Make sure every attribute with the same name and annotatable type has the same value type."
         )
         super().__init__(message)
 
--- a/hari_client/client/errors.py
+++ b/hari_client/client/errors.py
@@ -137,6 +137,6 @@ class AttributeValidationIdNotReusedError(Exception):
     ):
         message = (
             f"Found multiple ids {found_ids} for the same attribute name {attribute_name} with {annotatable_type=}."
-            + " Make sure every attribute with the same name and attribute type is using the same id."
+            + " Make sure every attribute with the same name and annotatable type is using the same id."
         )
         super().__init__(message)
--- a/hari_client/client/errors.py
+++ b/hari_client/client/errors.py
@@ -118,6 +118,19 @@ class AttributeValidationInconsistentListElementValueTypesError(Exception):
         super().__init__(message)
 
 
+class AttributeValidationInconsistentListElementValueTypesMultipleAttributesError(
+    Exception
+):
+    def __init__(
+        self, attribute_name: str, annotatable_type: str, found_value_types: list[str]
+    ):
+        message = (
+            f"Found multiple instances of attribute {attribute_name} with {annotatable_type=} with inconsistent list element value types {found_value_types}."
+            + " Make sure every instance of this attribute uses the same list element value types."
+        )
+        super().__init__(message)
+
+
 class AttributeValidationIdNotReusedError(Exception):
     def __init__(
         self, attribute_name: str, annotatable_type: str, found_ids: list[str]
--- a/hari_client/client/errors.py
+++ b/hari_client/client/errors.py
@@ -94,3 +94,36 @@ class BulkOperationAnnotatableIdMissing(Exception):
             "Its value should be a unique UUID for each item within the bulk."
         )
         super().__init__(message)
+
+
+class AttributeValidationInconsistentValueTypeError(Exception):
+    def __init__(
+        self, attribute_name: str, annotatable_type: str, found_value_types: list[str]
+    ):
+        message = (
+            f"Found multiple value types {found_value_types} for attribute {attribute_name} with {annotatable_type=}."
+            + " Make sure every attribute with the same name and attribute type has the same value type."
+        )
+        super().__init__(message)
+
+
+class AttributeValidationInconsistentListElementValueTypesError(Exception):
+    def __init__(
+        self, attribute_name: str, annotatable_type: str, found_value_types: list[str]
+    ):
+        message = (
+            f"Found multiple list element value types {found_value_types} for attribute {attribute_name} with {annotatable_type=}."
+            + " Make sure every list element has the same value type."
+        )
+        super().__init__(message)
+
+
+class AttributeValidationIdNotReusedError(Exception):
+    def __init__(
+        self, attribute_name: str, annotatable_type: str, found_ids: list[str]
+    ):
+        message = (
+            f"Found multiple ids {found_ids} for the same attribute name {attribute_name} with {annotatable_type=}."
+            + " Make sure every attribute with the same name and attribute type is using the same id."
+        )
+        super().__init__(message)
--- a/hari_client/models/validation/attribute_validation.py
+++ b/hari_client/models/validation/attribute_validation.py
@@ -1,0 +1,91 @@
+import numbers
+import typing
+
+from hari_client import errors
+from hari_client import models
+
+
+def validate_initial_attributes(
+    attributes: list[models.AttributeCreate],
+) -> None:
+    """Raises an error if any requirements for initial attributes aren't met.
+    Attribute requirements:
+    - attributes with the same name and annotatable type must
+        - have the same value type
+        - have the same id
+    - attributes with value type list, mustn't contain multiple value types in the list
+
+    Args:
+        attributes: List of attributes to validate.
+
+    Raises:
+        AttributeValidationInconsistentValueTypeError: If the value type for an attribute is inconsistent.
+        AttributeValidationInconsistentListElementValueTypesError: If the elements in a list attribute value have mixed value types.
+        AttributeValidationIdNotReusedError: If another attribute with the same name and attribute type has a different id.
+    """
+    grouped_by_attribute_type: dict[str, list[models.AttributeCreate]] = {
+        models.DataBaseObjectType.MEDIA: [],
+        models.DataBaseObjectType.MEDIAOBJECT: [],
+        models.DataBaseObjectType.INSTANCE: [],
+    }
+    for attribute in attributes:
+        grouped_by_attribute_type[attribute.annotatable_type].append(attribute)
+
+    for (
+        annotatable_type,
+        attributes_with_same_annotatable_type,
+    ) in grouped_by_attribute_type.items():
+        name_value_type_map: dict[str, str] = {}
+        name_id_map: dict[str, str] = {}
+
+        for attribute in attributes_with_same_annotatable_type:
+            # check value type
+            value_type = _get_value_type_for_comparison(attribute.value)
+            if attribute.name in name_value_type_map:
+                if name_value_type_map[attribute.name] != value_type:
+                    raise errors.AttributeValidationInconsistentValueTypeError(
+                        attribute_name=attribute.name,
+                        annotatable_type=annotatable_type,
+                        found_value_types=[
+                            name_value_type_map[attribute.name],
+                            value_type,
+                        ],
+                    )
+            else:
+                name_value_type_map[attribute.name] = value_type
+
+            # check list elements if value type is list
+            if value_type == "list":
+                found_list_element_value_types: set[str] = set()
+                for element in attribute.value:
+                    found_list_element_value_types.add(
+                        _get_value_type_for_comparison(element)
+                    )
+                if len(found_list_element_value_types) > 1:
+                    raise errors.AttributeValidationInconsistentListElementValueTypesError(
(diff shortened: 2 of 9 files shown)


hari_client/models/validation/attribute_validation.py
```diff
@@ -0,0 +1,91 @@
+import numbers
+import typing
+
+from hari_client import errors
+from hari_client import models
+
+
+def validate_initial_attributes(
```
katia-qm, line 8: All the validations you added are common between all types of attributes, so we could rename it to just `validate_attributes`

hari_client/models/validation/attribute_validation.py
```diff
@@ -0,0 +1,91 @@
...
+    attributes: list[models.AttributeCreate],
+) -> None:
+    """Raises an error if any requirements for initial attributes aren't met.
+    Attribute requirements:
+    - attributes with the same name and annotatable type must
+        - have the same value type
+        - have the same id
```
katia-qm, line 15: I totally get the reasoning here, but I'm still not sure whether we could put it as a fact.
In the docs for attributes we say that `attribute with the same **id** and annotatable type must be consistent`, so it's a contradiction

hari_client/models/validation/attribute_validation.py
```diff
@@ -0,0 +1,91 @@
...
+        attributes_with_same_annotatable_type,
+    ) in grouped_by_attribute_type.items():
+        name_value_type_map: dict[str, str] = {}
+        name_id_map: dict[str, str] = {}
+
+        for attribute in attributes_with_same_annotatable_type:
+            # check value type
+            value_type = _get_value_type_for_comparison(attribute.value)
+            if attribute.name in name_value_type_map:
+                if name_value_type_map[attribute.name] != value_type:
```
katia-qm, line 45: None is a valid value, go it can get mixed with other types. However i'm not sure if none is a valid value for initial attributes. Maybe we should add another validation here, but it's not there in the api. Anyway the validation function is general for all attributes now

hari_client/models/validation/attribute_validation.py
```diff
@@ -0,0 +1,91 @@
...
+                for element in attribute.value:
+                    found_list_element_value_types.add(
+                        _get_value_type_for_comparison(element)
+                    )
+                if len(found_list_element_value_types) > 1:
+                    raise errors.AttributeValidationInconsistentListElementValueTypesError(
+                        attribute_name=attribute.name,
+                        annotatable_type=annotatable_type,
+                        found_value_types=found_list_element_value_types,
+                    )
```
katia-qm, line 69: we should also check that types of list elements are the same as list we compare to

hari_client/client/errors.py
```diff
@@ -94,3 +94,36 @@ def __init__(self):
...
         super().__init__(message)
+
+
+class AttributeValidationInconsistentValueTypeError(Exception):
+    def __init__(
+        self, attribute_name: str, annotatable_type: str, found_value_types: list[str]
+    ):
+        message = (
+            f"Found multiple value types {found_value_types} for attribute {attribute_name} with {annotatable_type=}."
+            + " Make sure every attribute with the same name and attribute type has the same value type."
```
katia-qm, line 105:
```suggestion
            + " Make sure every attribute with the same name and annotatable type has the same value type."
```

katia-qm (changes requested):
Good job! I left come comments about missing validations.

The biggest concern I have is that we take name instead of id as a source of truth that goes against api implementation (and the docs), but we can discuss it

hari_client/client/errors.py
katia-qm: and everywhere else you find it

hari_client/models/validation/attribute_validation.py
Dens49: meaning it should be ok for `None` to be mixed in with anything?

hari_client/models/validation/attribute_validation.py
```diff
@@ -0,0 +1,134 @@
...
+import typing
+
+from hari_client import errors
+from hari_client import models
+
+
+def validate_attributes(
+    attributes: list[models.AttributeCreate],
+) -> None:
+    """Raises an error if any requirements for attributes aren't met.
```
katia-qm, line 11:
```suggestion
    """Raises an error if any requirements for attributes consistency aren't met.
```

hari_client/models/validation/attribute_validation.py
```diff
@@ -0,0 +1,134 @@
...
+    Basic attribute requirements:
+    - attributes with the same name and annotatable type must
+      - have the same id
+      - have the same value type
+    - attributes with value type list must
+      - contain only a single value type
+      - contain only the same value type as other attributes with the same name and annotatable type
```
katia-qm, line 18:
```suggestion
   - attributes with value type list must
      - have consistent element value types within the list
      - have consistent element value types across all values for the attributes with the same name and annotatable type
```

hari_client/models/validation/attribute_validation.py
```diff
@@ -0,0 +1,134 @@
...
+
+    Exceptions:
+      - The value `None` isn't taken into account when value type consistency is checked.
```
katia-qm, line 21:
```suggestion
    Exceptions:
      - The value `None` isn't taken into account when value type consistency is checked.
      - Numeric types (int, float) could be mixed
```

hari_client/models/validation/attribute_validation.py
```diff
@@ -0,0 +1,134 @@
...
+
+    Args:
+        attributes: List of attributes to validate.
+
+    Raises:
+        AttributeValidationInconsistentValueTypeError: If the value type for an attribute is inconsistent.
+        AttributeValidationInconsistentListElementValueTypesError: If the elements in a list attribute value have mixed value types.
+        AttributeValidationInconsistentListElementValueTypesMultipleAttributesError: If multiple list attributes have inconsistent element value types.
+        AttributeValidationIdNotReusedError: If another attribute with the same name and attribute type has a different id.
```
katia-qm, line 30:
```suggestion
        AttributeValidationInconsistentListElementValueTypesError: If the elements in a list attribute value have inconsistent value types.
        AttributeValidationInconsistentListElementValueTypesMultipleAttributesError: If multiple attributes with list value have inconsistent element value types.
        AttributeValidationIdNotReusedError: If multiple attributes with the same name and attribute type have different ids.
```

hari_client/upload/hari_uploader.py
```diff
@@ -368,6 +375,22 @@ def _handle_object_categories(self) -> None:
...
         self._assign_object_category_subsets()
 
+    def validate_all_attributes(self) -> None:
+        """Validates all attributes of medias and media objects.
+
+        Raises:
+            AttributeValidationInconsistentValueTypeError: If the value type for an attribute is inconsistent.
+            AttributeValidationInconsistentListElementValueTypesError: If the elements in a list attribute value have mixed value types.
+            AttributeValidationInconsistentListElementValueTypesMultipleAttributesError: If multiple list attributes have inconsistent element value types.
+            AttributeValidationIdNotReusedError: If another attribute with the same name and attribute type has a different id.
```
katia-qm, line 385: Do we have to duplicate the exceptions here? Maybe it's enough to have them in the validation function

hari_client/client/errors.py
```diff
@@ -94,3 +94,49 @@ def __init__(self):
...
+
+class AttributeValidationInconsistentValueTypeError(Exception):
+    def __init__(
+        self, attribute_name: str, annotatable_type: str, found_value_types: list[str]
+    ):
+        message = (
+            f"Found multiple value types {found_value_types} for attribute {attribute_name} with {annotatable_type=}."
+            + " Make sure every attribute with the same name and annotatable type has the same value type."
+        )
+        super().__init__(message)
```
katia-qm, line 107: Error messages seem a little bit confusing to me, but I also don't know how to rephrase it better.

The fact that we have to mention `with the same name and annotatable type` everywhere really kills me.

hari_client/models/validation/attribute_validation.py
```diff
@@ -0,0 +1,134 @@
...
+    }
+    for attribute in attributes:
+        grouped_by_annotatable_type[attribute.annotatable_type].append(attribute)
+
+    for (
+        annotatable_type,
+        attributes_with_same_annotatable_type,
+    ) in grouped_by_annotatable_type.items():
+        # to check the value type consistency of attribute.value
+        name_value_type_map: dict[str, str] = {}
```
katia-qm, line 45: Would be really nice to make it a little bit more modular, at least separate validation for list and non-list values to a separate function if possible. If we later add more validations the function will grow and make become a mess

tests/test_attribute_validations.py
```diff
@@ -0,0 +1,354 @@
...
+                id=_some_uuids[1],
+                name="b",
+                annotatable_id="media_object_1",
+                annotatable_type=models.DataBaseObjectType.MEDIAOBJECT,
+                value=0,
+            ),
+        ],
+    ],
+)
+def test_attribute_validation_inconsistent_value_type(attributes):
```
katia-qm, line 87: I didn't check all the tests yet, but maybe it would make sense to check the validation tests for the api to be sure we didn't forget any test cases

katia-qm (dismissed):
Great job, thanks for the effort! I know these validations are quite overwhelming, but it's worth it for sure.

I think consistency validation for annotatable type/attribute group/attribute type is still missing

hari_client/client/errors.py
Dens49: Think about it that way: as a user you'll hopefully only see those messages very rarely and so you're probably happy to see enough detail to explain the problem and the expectations.

hari_client/upload/hari_uploader.py
Dens49: yeah right, having them duplicated here doesn't seem to be helpful

Dens49:
I did a refactor of the validation code, which surfaced two bugs :partying_face: (they're fixed, see latest two commits).
I also added a few more test cases and tried to improve the explanations in the doc strings

hari_client/models/validation/attribute_validation.py
```diff
@@ -0,0 +1,164 @@
...
+                    ],
+                )
+        else:
+            self.name_list_element_value_type_map[
+                attribute.name
+            ] = attribute_list_element_value_type
+
+    def _check_attribute_id_usage(self, attribute: models.AttributeCreate) -> None:
+        if attribute.name in self.name_id_map:
+            if self.name_id_map[attribute.name] != str(attribute.id):
```
FBOBecker, line 146: hmm this is only an artifact because we actually use these merged Attribute Models for the client, right?
Technically, we store the name only once on the metadata, so kind of redundant here I assume.
Did we ever consider actually using the difference of attr metadata and value for the client?

hari_client/models/validation/attribute_validation.py
```diff
@@ -0,0 +1,164 @@
...
+
+    def validate_attributes(self) -> None:
+        """Validates that attributes in a list have consistent value types and ids."""
+        for attribute in self.attributes:
+            self._check_attribute_id_usage(attribute)
+
+            value_type = self._check_value_type(attribute)
+
+            # None values automatically pass the value type consistency check
+            if value_type != "NoneType":
```
FBOBecker, line 74: you could probably remove that part as it would be checked within the method either way

hari_client/models/validation/attribute_validation.py
```diff
@@ -0,0 +1,164 @@
...
+                found_value_types=found_list_element_value_types,
+            )
+
+        # check list element value type consistency across attributes
+        attribute_list_element_value_type = list(found_list_element_value_types)[0]
+        if attribute.name in self.name_list_element_value_type_map:
+            if (
+                self.name_list_element_value_type_map[attribute.name]
+                != attribute_list_element_value_type
+            ):
```
FBOBecker, line 130: Sorry, too unfamiliar with the client but: Do we allow for different attributes to be uploaded? E.g., brightness and contrast at the same time?

FBOBecker (approved):
Thanks for adding these. From what I can judge it is looking good. Just some questions here and there as I am not super familiar with the client (yet).

hari_client/models/validation/attribute_validation.py
```diff
@@ -0,0 +1,164 @@
...
+                    ],
+                )
+        else:
+            self.name_value_type_map[attribute.name] = value_type
+
+        return value_type
+
+    def _check_list_elements_value_types(
+        self, attribute: models.AttributeCreate, value_type: str
+    ) -> None:
```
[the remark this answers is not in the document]
Dens49: I used to be camp "only typehint what's returned explicitly", but Alvaro taught me we actually have the convention to also typehint `-> None` if there's no explicit return.
This method has one explicit return, but no return value, so always returns None

PR #56, opened 2024-12-13
Fix new client methods

katia-qm:
- fix fix get_visualisation_configs method _request params

### DIFF
--- a/hari_client/client/client.py
+++ b/hari_client/client/client.py
@@ -2133,8 +2133,8 @@ class HARIClient:
             list[models.VisualisationConfiguration]: A list of visualization configuration objects.
         """
         return self._request(
-            method="GET",
-            path=f"/datasets/{dataset_id}/visualisationConfigs",
+            "GET",
+            f"/datasets/{dataset_id}/visualisationConfigs",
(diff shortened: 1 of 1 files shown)


hari_client/client/client.py
```diff
@@ -2064,15 +2064,17 @@ def get_attribute_metadata(
...
-        """Returns all attributes of a dataset
+        """Returns all attribute metadata of a dataset.
 
         Args:
             dataset_id: The dataset id
-            archived: True if archived attributes should be returned
-            query: A query to filter attributes
+            archived: if True, only archived attribute metadata are returned
+                    if False, only non-archived attribute metadata are returned
+                    if None, all attribute metadata are returned
```
Dens49, line 2073: Let's try to think of a generic way of explaining the archived filter that could describe the behavior for every endpoint that has that arg. Maybe `if set to True or False, only archived or non-archived entities are returned. If set to None, the archived filter isn't applied.`

Dens49 (approved):

hari_client/client/client.py
katia-qm: done, does it look better now?

PR #57, opened 2024-12-16
include existing attributes when validating the number of attributes

katia-qm:
include existing attributes when validating the number of attributes

### DIFF
--- a/hari_client/upload/hari_uploader.py
+++ b/hari_client/upload/hari_uploader.py
@@ -408,6 +408,10 @@ class HARIUploader:
         Returns:
             HARIUploadResults | None: All upload results and summaries for the
             upload of medias and media_objects, or None if nothing was uploaded
+
+        Raises:
+            HARIUniqueAttributesLimitExceeded: If the number of unique attribute ids
+            exceeds the limit of MAX_ATTR_COUNT per dataset.
         """
 
         if len(self._medias) == 0:
--- a/hari_client/upload/hari_uploader.py
+++ b/hari_client/upload/hari_uploader.py
@@ -172,7 +172,7 @@ class HARIUploader:
         self._attribute_cnt: int = 0
         # TODO: this should be a dict[str, uuid.UUID] as soon as the api models are updated
         self._object_category_subsets: dict[str, str] = {}
-        self._unique_attribute_ids: set[uuid.UUID] = set()
+        self._unique_attribute_ids: set[str] = set()
 
     # TODO: add_media shouldn't do validation logic, because that expects that a specific order of operation is necessary,
     # specifically that means that media_objects and attributes have to be added to media before the media is added to the uploader.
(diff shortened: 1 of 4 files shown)


hari_client/upload/hari_uploader.py
```diff
@@ -386,10 +386,18 @@ def upload(
...
+            dataset_id=self.dataset_id
+        )
+        existing_attribute_ids = {attr.id for attr in existing_attr_metadata}
+        all_attribute_ids = existing_attribute_ids.union(self._unique_attribute_ids)
+
+        if len(all_attribute_ids) > MAX_ATTR_COUNT:
             raise HARIUniqueAttributesLimitExceeded(
                 f"You are trying to upload too many attributes for one dataset: {len(self._unique_attribute_ids)}, "
-                f"the limit is {MAX_ATTR_COUNT}. "
+                f"and there are already {len(existing_attr_metadata)} attributes uploaded. "
```
Dens49, line 398: nitpick: I forgot to raise that in the other PR: Consider defining the error message in the Exception class rather than where you raise the error and pass in the relevant variables.

I also find the error message to be quite confusing to be honest :grimacing: What do you think about rewriting it to be about the amount of unique attribute ids instead of just "attributes"?

tests/test_hari_uploader.py
```diff
@@ -1072,3 +1072,68 @@ def test_hari_uploader_unique_attributes_number_limit_error(
...
+
+    # Act + Assert
+    with pytest.raises(hari_uploader.HARIUniqueAttributesLimitExceeded) as e:
+        uploader.upload()
+    assert (
+        f"You are trying to upload too many attributes for one dataset: {new_attrs_number}, "
+        f"and there are already {len(mock_attribute_metadata)} attributes uploaded. "
+        f"The intended number of all attributes would be 1001, "
+        f"when the limit is {hari_uploader.MAX_ATTR_COUNT}. " in str(e.value)
+    )
```
Dens49, line 1139: nitpick: don't test the whole error message appearing exactly as defined. For the purpose of the test it's enough to simply check whether the correct exception type was raised. If you also want to test whether the error message contains the correct values, consider just checking whether the value appears or whether the correct args were passed to the exception object: https://stackoverflow.com/a/23514853

Dens49 (approved):
good job on the testing :+1: 
the hari_uploader utility doesn't officially support adding data to an already existing dataset, so the extra check doesn't hurt, but also isn't that necessary. It's a good reminder that we should have that functionality, though!

tests/test_hari_uploader.py
katia-qm: Great advice, thanks, I was looking for that, of course I should have defined error arguments

hari_client/upload/hari_uploader.py
katia-qm: agree

Dens49 (approved):
thanks, good changes!
final nitpick: the `HARIUniqueAttributesLimitExceeded` exception should be mentioned in the `Raises` section of the hari_uploader::upload method

katia-qm:
> thanks, good changes! final nitpick: the `HARIUniqueAttributesLimitExceeded` exception should be mentioned in the `Raises` section of the hari_uploader::upload method

Glaz almaz!

PR #58, opened 2025-01-10
Add data upload scripts

Emprime:
I added wrappings and scripts for data upload. I specifically showed how to upload a dataset with one line, a coco-like dataset, and a dataset with its own annotation attributes. 

Added also an upload option which checks for already uploaded images and prevents them from reuploading.


Emprime:
There is one TODO open: # TODO what is a good default user group, how should a customer select it

What is a good user_group during dataset upload?

Emprime:
Moreover, It would be great to add an integration test on a dedicated server instance, @Dens49 you mentioned once someone is working on it, is there an update on this topic?

Emprime:
Comments to this topic:

"Yeah sure, but what if the upload breaks in between the batch upload? The client uploads the Media first to achieve the Hari UUID for that media. I'm not quite sure if it will first upload the whole batch of Medias and after that start the mediaObject creation or if that also happens in parallel. But anyways: if the Upload process crashes when uploading a mediaObject, we will probably have some mediaObjects already created for one particular Media but some not. Same with the attributes

So there are several things we need to take care of:
Check the back reference of the mediaObjects as well and if not: upload it
Fetch the Attribute UUIDs from the server prior to the upload process itself and keep this in a lookup because the client needs to provide the Hari-internal UUID for the particular attribute key
check if this attribute already exists for that media/mediaObject and if not: upload it
There are probably other things to check/handle... that's just what I recognized when uploading data"

Maybe it is needed to discuss what is missing here and add it. For all of the mentioned problems I have at least partially worked on a solution.

Emprime:
The raised Issue of state awareness and user groups has been addressed in newer version here: https://github.com/quality-match/hari-client/pull/64

@Dens49 PR 64 is build on this non-reviewed branch, This PR contains a lot of methodology which is not touched by the new branch, however some scripts like upload are modified in the PR 64 to use the State Aware Uploader instead of the current default one

scripts/upload_generic_dataset.py
```diff
@@ -0,0 +1,72 @@
...
+
+    # load hari client
+    config: Config = Config(_env_file=".env")
+    hari: HARIClient = HARIClient(config=config)
+
+    # create dataset
+    dataset_id = check_and_create_dataset(
+        hari=hari, dataset_name=dataset_name, user_group=None, is_anonymized=True
+    )
+    # TODO what is a good default user group, how should a customer select it
```
katia-qm, line 49: this is how we do it in the quickstart, and in the docs we explain how to find out which one the user needs

```suggestion
    dataset_id = check_and_create_dataset(
        hari=hari, dataset_name=dataset_name, user_group="CHANGEME", is_anonymized=True
    )
    # TODO what is a good default user group, how should a customer select it
```

hari_client/utils/upload.py
```diff
@@ -0,0 +1,194 @@
...
+            ]
+        )
+        if jobs_are_still_running:
+            print(f"waiting for metadata_rebuild jobs to finish, {job_statuses=}")
+            time.sleep(10)
+
+    print(f"metadata_rebuild jobs finished with status {job_statuses=}")
+
+
+def check_and_create_dataset(
```
katia-qm, line 53: maybe `get_or_create_dataset`? it's a more common python convention

hari_client/utils/upload.py
```diff
@@ -0,0 +1,194 @@
...
+    """
+    Check if a subset with the given name exists within a dataset. If not, create it.
+
+    :param hari: An instance of HARIClient to interact with the HARI API.
+    :param dataset_id: The UUID of the dataset.
+    :param subset_name: The name of the subset to check or create.
+    :param subset_type: The type of the subset to create if it doesn't exist.
+    :return: A tuple containing:
+        - The UUID of the found or created subset.
+        - A boolean indicating whether the subset was reused (True) or newly created (False).
```
katia-qm, line 96:
```suggestion
        - A boolean indicating whether the subset with the given name already exists.
```

hari_client/utils/upload.py
```diff
@@ -0,0 +1,194 @@
...
+    )
+
+    uploaded_medias = hari.get_medias(dataset_id)
+    uploaded_back_references = [m.back_reference for m in uploaded_medias]
+
+    elements_to_upload = 0
+    for media in medias:
+        if media.back_reference not in uploaded_back_references:
+            uploader.add_media(media)
+            elements_to_upload += 1
```
katia-qm, line 148: I'm not sure if we should still keep it here (for the main repo and old uploader version I mean)

hari_client/utils/upload.py
```diff
@@ -0,0 +1,194 @@
...
+        print("Found subset with id:", new_subset_id)
+
+        return new_subset_id, True
+
+
+def check_and_upload_dataset(
+    hari: HARIClient,
+    dataset_id: uuid.UUID,
+    object_categories: list[str],
+    medias: list[Media],
```
katia-qm, line 120: `medias: list[HARIMedia],`

with `from hari_client.upload.hari_uploader import HARIMedia`

hari_client/utils/upload.py
```diff
@@ -0,0 +1,194 @@
...
+        new_subset_id = subset.id
+        print("Found subset with id:", new_subset_id)
+
+        return new_subset_id, True
+
+
+def check_and_upload_dataset(
+    hari: HARIClient,
+    dataset_id: uuid.UUID,
+    object_categories: list[str],
```
katia-qm, line 119: set[str] here and should be optional for media datasets

scripts/upload_dataset_with_own_attributes.py
```diff
@@ -0,0 +1,249 @@
...
+    #   one answer are written
+
+    answer_frequencies = {
+        image_path: {
+            cat_name: answer_cnt
+            for cat_name, answer_cnt in frequency.items()
+            if answer_cnt > 0
+        }
+        for image_path, frequency in (df_answers_pivot.to_dict(orient="index").items())
+    }
```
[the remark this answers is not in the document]
Emprime: Good Point I changed it

scripts/upload_dataset_with_own_attributes.py
```diff
@@ -0,0 +1,249 @@
...
+        attribute = hc.hari_uploader.HARIAttribute(
+            value=majority_vote,
+            frequency=frequency,
+            cant_solves=0,
+            repeats=avg_number_annotions,
+            **annotatable_attribute
+        )
+        media_object.add_attribute(attribute)
+        media_object.set_object_category_subset_name(majority_vote)
+        media.add_media_object(media_object)
```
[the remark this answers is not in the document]
Emprime: The uploaded data is an image classification dataset thus no Bounding boxes exist :D

hari_client/utils/upload.py
```diff
@@ -0,0 +1,194 @@
...
+        print("Dataset created with id:", new_dataset.id)
+        return new_dataset.id
+    else:
+        dataset_id = datasets[dataset_names.index(dataset_name)].id
+        print("Found existing dataset with id:", dataset_id)
+
+        return dataset_id
+
+
+def check_and_create_subset_for_all(
```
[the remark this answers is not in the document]
Emprime: Done

scripts/upload_dataset_with_own_attributes.py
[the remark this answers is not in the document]
katia-qm: Can you also remove the comment above saying that we only write categories with at least 1 answer please

scripts/upload_dataset_with_own_attributes.py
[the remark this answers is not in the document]
katia-qm: Do you think it's a good example then? I'd expect as an example data with real bboxes, the one here for the whole images is more a workaround

hari_client/utils/upload.py
```diff
@@ -0,0 +1,194 @@
...
+
+def trigger_and_display_metedata_update(
+    hari: HARIClient, dataset_id: uuid.UUID, subset_id: uuid.UUID | None
+):
+    """
+    Trigger and wait for the metadata update jobs to finish, then print their statuses.
+
+    :param hari: An instance of HARIClient to interact with the HARI API.
+    :param dataset_id: The UUID of the dataset for which the metadata should be rebuilt.
+    :param subset_id: The UUID of a specific subset, or None to rebuild metadata for the entire dataset.
```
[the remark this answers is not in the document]
Emprime: Updated

hari_client/utils/upload.py
Emprime: Depends a bit on what happens with the state-aware uploader...
I would argue it is better then no support at the moment but it is borderline.

hari_client/utils/upload.py
```diff
@@ -0,0 +1,194 @@
...
+            print(f"media upload details: {upload_results.medias.results}")
+
+    else:
+        print("WARNING: No data for upload specified which is not already uploaded.")
+
+    new_subset_id, reused = check_and_create_subset_for_all(
+        hari, dataset_id, new_subset_name, subset_type
+    )
+
+    if reused:
```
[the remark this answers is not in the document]
Emprime: agreed, changed it

hari_client/utils/upload.py
Emprime: Done

hari_client/utils/upload.py
Emprime: Done

hari_client/utils/upload.py
katia-qm: I'd take it out from here for now, leaving it here will provide partly state awareness (we are not checking media objects here for example). You and cs will be using state aware uploader anyway I guess.

scripts/upload_generic_dataset.py
```diff
@@ -0,0 +1,78 @@
...
+        file_path=local_path,
+        back_reference=image_url,
+        media_type=models.MediaType.IMAGE,
+    )
+
+    # upload image to dataset
+    check_and_upload_dataset(
+        hari=hari,
+        dataset_id=dataset_id,
+        object_categories=[],
```
katia-qm, line 74:
```suggestion
```

scripts/upload_dataset_with_own_attributes.py
```diff
@@ -0,0 +1,252 @@
...
+import json
+import os
+import uuid
+from pathlib import Path
+
+import pandas as pd
+from PIL import Image
+from sklearn.utils import Bunch
+
+import hari_client as hc
```
katia-qm, line 11: can we replace this to be consistent with other scripts removing hc?

let's just directly import what you need

```
from hari_client import Config
from hari_client import hari_uploader
from hari_client import HARIClient
from hari_client import models
```

scripts/upload_dataset_with_own_attributes.py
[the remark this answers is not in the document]
katia-qm: Yeah not sure, you not necessarily have to provide a real dataset here. It's a good example but I don't want to encourage users to to upload whole images as objects for aint. Let's ask Dennis' opinion

scripts/README.md
```diff
@@ -11,11 +11,32 @@ Please be aware that this folder is currently under construction and may include
...
+  - Example: `python upload_coco_like_dataset.py --dataset_name ms_coco_2017_val --image_directory /Path/to/my/data/mscoco/val2017 --annotations_file /Path/to/my/data/mscoco/annotations/instances_val2017.json --user_group USER`
+- `upload_dataset_with_own_attributes` Uploads a complete dataset with own annotation attributes e.g. for training AINTs. This example script uses the public available [Data-Centric Image Classification Benchmark Data](https://zenodo.org/records/8115942).
+  - Call with `python upload_dataset_with_own_attributes.py --root_directory <Path> --source_dataset_name <Name> --question "<Question describing task> --attribute_name <Name of Attribute> --user_group <USER_GROUP>"`
+  - Example: `python upload_dataset_with_own_attributes.py --root_directory /Path/to/my/data/DCIC --source_dataset_name QualityMRI --question "How is the quality of the image?" --attribute_name image_quality --user_group USER`
+- `trigger_metadata_rebuild` Triggers the metadata rebuild like crop generation, missing thumbnails, wrong histogram calculations.
```
katia-qm, line 31:
```suggestion
- `trigger_metadata_rebuild` Triggers the metadata rebuild like creating default visualizations for thumbnails and crops, updating the histograms for advanced filtering, etc
```

katia-qm (commented):
Looks good to me overall, thanks for the effort!

Let's discuss open points that were left

scripts/upload_yolo_like_dataset.py
```diff
@@ -0,0 +1,264 @@
...
+                    # Convert normalized coords -> absolute pixel coords
+                    x_center = x_norm * width
+                    y_center = y_norm * height
+                    w_abs = w_norm * width
+                    h_abs = h_norm * height
+
+                    # We label the class as "class_{ID}"
+                    # Convert the class index to its string label:
+                    if str(class_idx) in class_names:
+                        cat_name = class_names[str(class_idx)]
```
katia-qm, line 108: nitpick: can we have `category_name` here as well please?

scripts/upload_generic_dataset.py
```diff
@@ -0,0 +1,78 @@
+import argparse
```
katia-qm, line 1: is this script useful if we don't use the state aware uploader?

scripts/upload_dataset_with_own_attributes.py
```diff
@@ -0,0 +1,252 @@
+import argparse
```
katia-qm, line 1: can we name is ..._own_annotation_attributes?

hari_client/utils/upload.py
```diff
@@ -0,0 +1,202 @@
...
+        dataset_name: The name of the dataset to check or create.
+        user_group: The user group under which the dataset should be created.
+        is_anonymized: Whether the dataset should be created with anonymized data.
+    Returns:
+         uuid: The UUID of the found or created dataset.
+    """
+    datasets = hari.get_datasets()
+    dataset_names = [dataset.name for dataset in datasets]
+
+    assert user_group is not None, "User group is required."
```
Dens49, line 71: Please don't use assert statements in the package source code. We don't control the runtime environment of whichever scripts that might end up using this code. Assert statements can be disabled in the interpreter settings. Rather throw an exception if you want to explicitly handle problematic input values.

Dens49 (commented):
:ok_hand:

Dens49 (approved):
approvado

katia-qm (approved):

PR #62, opened 2025-01-28
Add segmentation support

Schmaexxi:
# Background

1773466013

We are about to support segmentations in HARI. Client should follow API

# Changes
- added support for segmentation of media objects
- added model RLE-compressed bitmasks

## Monday
https://quality-match.monday.com/boards/1606259534/pulses/1773630492


katia-qm (approved):
Splendid!

PR #66, opened 2025-02-11
Restrict attribute possible_values to be a list of strings

katia-qm:
1795640538

Restrict attribute possible_values to be a list of strings


FBOBecker (commented):
looking good, thanks!

FBOBecker (approved):

PR #67, opened 2025-02-13
Update get datasets method with pagination and query, add get dataset count method

katia-qm:
Update get datasets method with pagination and query, add get dataset count method

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 ## [major.minor.patch] - DD-MM-YYYY
 
+### New Features
+
+- add skip, limit, sort, query and archived parameters to get_datasets method [PR#67](https://github.com/quality-match/hari-client/pull/67)
+- add get_datasets_count method [PR#67](https://github.com/quality-match/hari-client/pull/67)
+
 ## [3.1.0] - 14-01-2025
 
 ### New Features
--- a/hari_client/client/client.py
+++ b/hari_client/client/client.py
@@ -581,7 +581,7 @@ class HARIClient:
             query: query parameters to filter the datasets
             sort: sorting parameters to sort the datasets
             name_filter: filter by dataset name
-            archived: whether to include archived datasets (default: False)
+            archived: if true, return only archived datasets; if false (default), return non-archived datasets.
 
         Returns:
             A list of datasets
@@ -609,7 +609,7 @@ class HARIClient:
             visibility_statuses: Visibility statuses of the returned datasets
             query: query parameters to filter the datasets
             name_filter: filter by dataset name
-            archived: whether to consider archived datasets (default: False)
+            archived: if true, count only archived datasets; if false (default), count non-archived datasets.
 
         Returns:
             The number of datasets
--- a/hari_client/client/client.py
+++ b/hari_client/client/client.py
@@ -949,7 +949,7 @@ class HARIClient:
 
         Args:
             dataset_id: The dataset id
-            archived: Whether to get archived media
+            archived: if true, return only archived medias; if false (default), return non-archived medias.
             presign_medias: Whether to presign medias
             limit: The number of medias tu return
             skip: The number of medias to skip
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 ## [major.minor.patch] - DD-MM-YYYY
 
+### New Features
```
Schmaexxi, line 3: would be nice to already specify the new version based on the changes in this PR so the next person would know which version increase was intended

hari_client/client/client.py
```diff
@@ -584,6 +596,34 @@ def get_datasets(
...
+        name_filter: str | None = None,
+        archived: bool | None = False,
+    ) -> int:
+        """
+        Returns dataset count for the user.
+        Args:
+            visibility_statuses: Visibility statuses of the returned datasets
+            query: query parameters to filter the datasets
+            name_filter: filter by dataset name
+            archived: whether to include archived media_objects (default: False)
```
Schmaexxi, line 612:
```suggestion
            archived: if set, filters by archived dataset, specifically (default: False)
```

hari_client/client/client.py
```diff
@@ -564,12 +564,24 @@ def get_datasets(
...
 
         Args:
             subset: Return also subsets. If False, returns only parent datasets
             visibility_statuses: Visibility statuses of the returned datasets
+            limit: limit the number of datasets returned
+            skip: skip the number of datasets returned
+            query: query parameters to filter the datasets
+            sort: sorting parameters to sort the datasets
+            name_filter: filter by dataset name
+            archived: whether to include archived media_objects (default: False)
```
Schmaexxi, line 584:
```suggestion
            archived: if set, filters by archived dataset, specifically (default: False)
```

Schmaexxi (commented):
Thanks for the update! Could you update the docstrings with the suggestions, please?

CHANGELOG.md
katia-qm: I thought like it was fixes - patch, features - minor and breaking changes - major implicitly

Schmaexxi: I thought of it as a simplification for the next person to know what effect your changes would make on the next release (major/minor/patch).
Probably not a good idea if we dont have a convention for it, sorry - never mind

hari_client/client/client.py
katia-qm: Thanks a lot of catching this! It means that it's also wrong in the api, I blindly copied it from there

I rephrased it a little bit to be consistent with other docstings in the client

CHANGELOG.md
katia-qm: I also think it's a good idea, we just discussed it with Fabio yesterday as well, but we should come up with a convention

hari_client/client/client.py
Schmaexxi: Hmm, my main point wasn't that it wasn't specifying the correct object (dataset instead of media objects) but rather how it works.
It is an exclusive filter. It will either return all unarchived datasets if the parameter is false, or all archived ones if true @katia-qm

katia-qm: Hmm actually not really

False means only include not-archived items
True means only include archived ones
None means include all items

I remember we had this conversation in December and wanted to update the docsting to something like this everywhere

archived: if True, archived datasets will be returned, 
        if False, only non-archived datasets will be returned
        if None, all datasets will be returned

Schmaexxi: If that's what we want to achieve, I don't think it works or I'd like to see a test for it
https://github.com/quality-match/goliat/blob/b54b700c444fae971cde049943e8a7bd7c9b76e7/hari/hari/backend/routers/dataset.py#L403
it's never `None`

katia-qm:
Does the update look good to you @Schmaexxi? If so I can update it for all other endpoints

Schmaexxi:
> Does the update look good to you @Schmaexxi? If so I can update it for all other endpoints

Perfect, thank you!

PR #68, opened 2025-02-14
Remove client methods for obsolete api endpoints

Schmaexxi:
## Background
We changed the way thumbnails and crops are created in the backend. As a result, the API-endpoints
- /datasets/{dataset_id}/thumbnails
- /datasets/{dataset_id}/crops
were removed

## Changes
Remove `trigger_thumbnails_creation_job` and `trigger_crops_creation_job`

## Monday
https://quality-match.monday.com/boards/1606259534/pulses/1797705485

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## [major.minor.patch] - DD-MM-YYYY
 
+## [4.0.0] - TBD
+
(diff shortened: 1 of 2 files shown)


FBOBecker (approved):
Thanks for the clean up! 🙌

PR #72, opened 2025-02-21
Release/v3.2.0

katia-qm:
- added `compute_auto_attributes` param to `trigger_metadata_rebuild_job` 
- add `skip`, `limit`, `sort`, `query`, `name_filter` and `archived` parameters to `get_datasets` method 
- add `get_datasets_count` method 
- restricted attribute `possible_values` to be a list of strings


mylius (approved):
Thanks for taking care of the release!

PR #73, opened 2025-02-25
added low level support for external media sources

Dens49:
adds support for sending an external media source configuration when creating a dataset and added new endpoint to get an external media source, together with all necessary api models.

I want to use the base branch `external_media_sources` as a feature branch. A second PR with the updated HARIUploader flow will follow after this one.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 ## [major.minor.patch] - DD-MM-YYYY
 
+## [3.3.0] - DD-MM-YYYY
+
+### New features
+
+#### Support for external media sources
+
+- added support for defining external media sources when creating a dataset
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
 ## [major.minor.patch] - DD-MM-YYYY
 
+## [3.3.0] - DD-MM-YYYY
+
+### New features
+
+#### Support for external media sources
```
katia-qm, line 7: we didn't add subsections here before, but I don't mind

katia-qm (approved):
Everything looks right, thanks!

CHANGELOG.md
Dens49: oh right. Felt like it connects the dots :)

PR #74, opened 2025-02-25
hari-client and hari_uploader utility update for supporting external media sources

Dens49:
1777395874

add support for skipping media file upload in create_medias and hari_uploader when all medias have a set media_url. HARIMedia validates whether the set media_url is valid as a url for an media in a dataset that uses an external media source

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,16 @@
 - added support for defining external media sources when creating a dataset
   - new field `external_media_source` in the `create_dataset` method [PR#73](https://github.com/quality-match/hari-client/pull/73)
 - added new endpoint `get_external_media_source` [PR#73](https://github.com/quality-match/hari-client/pull/73)
+- added new arg to client method `create_medias` [PR#74](https://github.com/quality-match/hari-client/pull/74)
+  - `with_media_file_upload` (default: `True`). Set this to `False` if you want to skip the upload of media files.
+  - the method will then only send the `BulkMediaCreate` instance to HARI.
+- added format validations for `HARIMedia` field `media_url` [PR#74](https://github.com/quality-match/hari-client/pull/74)
+  - the field was previously set automatically through the process of uploading media files.
+  - in order to work with a dataset that uses an external media source, you have to set the field explicitly.
+- updated usage of HARIUploader utility [PR#74](https://github.com/quality-match/hari-client/pull/74)
+  - when using an external media source, make sure to only set the `media_url` field of a `HARIMedia` and don't set the `file_path`.
+  - if all medias have a set `media_url` field, then media files will not be uploaded to HARI, but they're only setup with the api.
+  - if all medias have a set `file_path` field, then the uploader utility behaves as before and will upload media files to HARI.
 
 ## [3.2.0] - 25-02-2025
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
   - new field `external_media_source` in the `create_dataset` method [PR#73](https://github.com/quality-match/hari-client/pull/73)
 - added new endpoint `get_external_media_source` [PR#73](https://github.com/quality-match/hari-client/pull/73)
 - added new arg to client method `create_medias` [PR#74](https://github.com/quality-match/hari-client/pull/74)
-  - `with_media_file_upload` (default: `True`). Set this to `False` if you want to skip the upload of media files. This way the upload will only create medias in HARI without uploading media files to QM storage.
+  - `with_media_files_upload` (default: `True`). Set this to `False` if you want to skip the upload of media files. This way the upload will only create medias in HARI without uploading media files to QM storage.
 - added format validations for `HARIMedia` field `media_url` [PR#74](https://github.com/quality-match/hari-client/pull/74)
   - the field was previously set automatically through the process of uploading media files.
   - in order to work with a dataset that uses an external media source, you have to set the field explicitly.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,15 +10,13 @@
   - new field `external_media_source` in the `create_dataset` method [PR#73](https://github.com/quality-match/hari-client/pull/73)
 - added new endpoint `get_external_media_source` [PR#73](https://github.com/quality-match/hari-client/pull/73)
 - added new arg to client method `create_medias` [PR#74](https://github.com/quality-match/hari-client/pull/74)
-  - `with_media_file_upload` (default: `True`). Set this to `False` if you want to skip the upload of media files.
-  - the method will then only send the `BulkMediaCreate` instance to HARI.
+  - `with_media_file_upload` (default: `True`). Set this to `False` if you want to skip the upload of media files. This way the upload will only create medias in HARI without uploading media files to QM storage.
 - added format validations for `HARIMedia` field `media_url` [PR#74](https://github.com/quality-match/hari-client/pull/74)
   - the field was previously set automatically through the process of uploading media files.
   - in order to work with a dataset that uses an external media source, you have to set the field explicitly.
-- updated usage of HARIUploader utility [PR#74](https://github.com/quality-match/hari-client/pull/74)
-  - when using an external media source, make sure to only set the `media_url` field of a `HARIMedia` and don't set the `file_path`.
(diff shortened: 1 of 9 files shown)


hari_client/upload/hari_uploader.py
```diff
@@ -451,6 +514,7 @@ def upload(
             )
 
         self.validate_all_attributes()
+        self._prepare_media_upload()
```
katia-qm, line 517: not sure, but maybe it should go before attribute validation?

hari_client/client/client.py
```diff
@@ -832,14 +832,18 @@ def create_media(
         )
 
     def create_medias(
-        self, dataset_id: uuid.UUID, medias: list[models.BulkMediaCreate]
+        self,
+        dataset_id: uuid.UUID,
+        medias: list[models.BulkMediaCreate],
+        with_media_file_upload: bool = True,
```
katia-qm, line 838: nitpick: maybe with_media_files_upload? sounds better to me

tests/test_hari_uploader.py
```diff
@@ -1135,3 +1150,147 @@ def test_hari_uploader_unique_attributes_number_limit_error_with_existing_attrib
...
+            media_type=models.MediaType.IMAGE,
+            file_path="path/to/image_2.jpg",
+        ),
+    ]
+    uploader.add_media(*medias)
+
+    # Act + Assert
+    with pytest.raises(hari_uploader.HARIMediaUploadError):
+        uploader._prepare_media_upload()
+    # assert uploader._with_media_file_upload
```
katia-qm, line 1296: I guess we don't care:)

CHANGELOG.md
```diff
@@ -9,6 +9,16 @@
 - added support for defining external media sources when creating a dataset
   - new field `external_media_source` in the `create_dataset` method [PR#73](https://github.com/quality-match/hari-client/pull/73)
 - added new endpoint `get_external_media_source` [PR#73](https://github.com/quality-match/hari-client/pull/73)
+- added new arg to client method `create_medias` [PR#74](https://github.com/quality-match/hari-client/pull/74)
+  - `with_media_file_upload` (default: `True`). Set this to `False` if you want to skip the upload of media files.
+  - the method will then only send the `BulkMediaCreate` instance to HARI.
```
katia-qm, line 14: maybe something like

```suggestion
  - `with_media_file_upload` (default: `True`). Set this to `False` if you want to skip the upload of media files. This way the upload will only create medias in HARI without uploading media files to QM storage.
```

CHANGELOG.md
```diff
@@ -9,6 +9,16 @@
...
+- added format validations for `HARIMedia` field `media_url` [PR#74](https://github.com/quality-match/hari-client/pull/74)
+  - the field was previously set automatically through the process of uploading media files.
+  - in order to work with a dataset that uses an external media source, you have to set the field explicitly.
+- updated usage of HARIUploader utility [PR#74](https://github.com/quality-match/hari-client/pull/74)
+  - when using an external media source, make sure to only set the `media_url` field of a `HARIMedia` and don't set the `file_path`.
+  - if all medias have a set `media_url` field, then media files will not be uploaded to HARI, but they're only setup with the api.
+  - if all medias have a set `file_path` field, then the uploader utility behaves as before and will upload media files to HARI.
```
katia-qm, line 21: I'm afraid it can be error prone. I think ideally we should provide a way for the user to "mark" the upload for the external source usage, and based on that do our validations. What do you think?

hari_client/upload/hari_uploader.py
```diff
@@ -170,6 +172,43 @@ def empty_back_reference(cls, v: str) -> str:
             )
         return v
 
+    @pydantic.field_validator("media_url")
+    @classmethod
+    def media_url_is_valid(cls, v: str) -> str:
```
katia-qm, line 177: is this something that is also defined in the api?

katia-qm (commented):
Good job, super clean, great tests!

I have some concerns about how we make a decision whether the upload is for external source, I believe we should do it the other way round, let's discuss it

katia-qm (commented):
Ah and would be nice to have a test for the whole upload (with mocks) and check that we don't upload medias in the end

hari_client/upload/hari_uploader.py
Dens49: moved it to the beginning of the upload method

CHANGELOG.md
Dens49: I used the VSCode github PR integration to apply the suggestion. This somehow didn't commit that explicitly, though

katia-qm: I think doing it from the ide just applies it locally

tests/test_client.py
```diff
@@ -23,6 +23,22 @@ def test_create_medias_with_missing_file_paths(test_client):
         client.create_medias(dataset_id="1234", medias=[media_create])
 
 
+def test_create_medias_with_missing_media_urls(test_client):
```
katia-qm, line 26: nice-to-have: I'd also expect here a non-failing test for external source case where we check that _upload_media_files_with_presigned_urls was never called

katia-qm (changes requested):
I just saw you fixed the error in create_medias in the client, so nothing major left, left some nice-to-have comments

I think we missed allowing the user to upload a single media with external source

hari_client/client/client.py
```diff
@@ -832,14 +832,18 @@ def create_media(
         )
 
     def create_medias(
-        self, dataset_id: uuid.UUID, medias: list[models.BulkMediaCreate]
+        self,
```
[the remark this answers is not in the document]
Dens49: I think it's so rarely used that I forgot this even exists. Thanks, will add that!

hari_client/upload/hari_uploader.py
```diff
@@ -88,6 +90,10 @@ class HARIMediaObjectUnknownObjectCategorySubsetNameError(Exception):
     pass
 
 
+class HARIMediaValidationError(Exception):
```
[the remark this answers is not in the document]
Dens49: in the end it's a context-dependent validation on the fileds that are set on the media. The context is whether or not an external media source is used, which is given in the two instances where the exception is thrown

hari_client/upload/hari_uploader.py
```diff
@@ -196,6 +239,7 @@ def __init__(
         self.client: HARIClient = client
         self.dataset_id: uuid.UUID = dataset_id
         self.object_categories = object_categories or set()
+        self._dataset: models.DatasetResponse | None = None
```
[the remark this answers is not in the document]
Dens49: yeah right. keeping it as a field adds unnecessary state that we're not doing anything with at the moment

hari_client/upload/hari_uploader.py
```diff
@@ -415,6 +468,47 @@ def validate_all_attributes(self) -> None:
...
+                raise HARIMediaValidationError(
+                    f"Dataset with id {self.dataset_id} uses an external media source, but not all medias {medias_with_media_url_cnt}/{len(self._medias)} have a media_url set. Make sure to set their media_url."
+                )
+        else:
+            if medias_with_file_path_cnt == len(self._medias):
+                self._with_media_files_upload = True
+            else:
+                raise HARIMediaValidationError(
+                    f"Dataset with id {self.dataset_id} requires media files to be uploaded, but not all medias {medias_with_file_path_cnt}/{len(self._medias)} have a file_path set. Make sure to set their file_path."
+                )
```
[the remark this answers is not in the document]
Dens49: I think the behavior of both versions is the same, but yours is easier to read :+1:

hari_client/upload/hari_uploader.py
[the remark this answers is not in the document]
katia-qm: agreed)

hari_client/client/client.py
```diff
@@ -787,6 +787,8 @@ def create_media(
         visualisations: list[models.VisualisationUnion] | None = None,
         subset_ids: set[str] | None = None,
         metadata: models.ImageMetadata | models.PointCloudMetadata | None = None,
+        media_url: str | None = None,
+        with_media_files_upload: bool = True,
```
katia-qm, line 791: we have to exclude `with_media_files_upload` from the json body here, otherwise it breaks

katia-qm (approved):
Great job, thanks for the changes and great tests!

PR #81, opened 2025-03-17
redundant (State aware uploader)

katia-qm:
The current uploader is a single process that pushes everything (media, media objects, attributes) in one go and relies on manual filtering to avoid duplicates via a “backreference” check. It also requires extra steps to retrieve old attribute IDs before uploading, forcing an all-or-nothing approach. This makes incremental updates (e.g., only uploading media objects) difficult and introduces the risk of creating duplicate entries when re-running uploads.

In the new system, a “state_aware_uploader.py” will replace the old uploader. It checks for existing items (media, media objects, attributes) and updates or creates them accordingly, allowing for partial uploads (only media or only attributes) and better handling of duplicates through backreference checks. The process optionally includes a thorough (but time-consuming) validation of attribute existence. Additionally, placeholder “mockup” classes can be used to stage existing media or media_objects without performing a full upload, making the process more flexible and efficient when you only want to upload media_objects or attributes.


katia-qm:
TODO: conflict status issue

PR #83, opened 2025-04-07
Fix small bugs for DS

katia-qm:


hari_client/utils/upload.py
```diff
@@ -55,7 +55,11 @@ def trigger_and_display_metadata_update(
 
 
 def get_or_create_dataset(
-    hari: HARIClient, dataset_name: str, user_group: str, is_anonymized: bool
+    hari: HARIClient,
+    dataset_name: str,
+    user_group: str,
+    is_anonymized: bool,
+    external_media_source: models.ExternalMediaSourceAPICreate,
```
Dens49, line 62: I think this should be optional for this method, especially because it's "get or create". Passing the external media source shouldn't be required for getting a dataset

Dens49 (changes requested):

hari_client/utils/upload.py
katia-qm: ouch, of course

Dens49 (approved):
:ok_hand:

PR #85, opened 2025-04-09
replace hari_api_base_url default with new api domain

FBOBecker:
As the primary HARI API domain was replaced by "api.hari.quality-match.com", we should set this as the new default.
Note that the old domain is still supported for some time to maintain compatibility.


Dens49 (approved):
thanks!

PR #88, opened 2025-04-23
added user group as argument to update ds method

BenjaminBentner:
The user group was missing as a possible argument in the update_dataset method.


Dens49 (approved):
:pray:

PR #100, opened 2025-06-30
[Patch] Add fake responses to keep server responses consistent when s…

Schmaexxi:
# Background
The upload summary and its statistics were generated from all the individual responses from the server.
That meant skipping some uploads of media_objects or attributes would reduce the number of total responses from the server because we did not even try to upload it.

[Aggregation of upload statistics](https://github.com/quality-match/hari-client/blob/75ac9c10524d2e2b4e288c7eff697bd22a5e99d8/hari_client/upload/hari_uploader.py#L1222-L1224)
[Uploader results](https://github.com/quality-match/hari-client/blob/75ac9c10524d2e2b4e288c7eff697bd22a5e99d8/hari_client/upload/hari_uploader.py#L677-L679)

[media objects are being excluded from upload](https://github.com/quality-match/hari-client/blob/75ac9c10524d2e2b4e288c7eff697bd22a5e99d8/hari_client/upload/hari_uploader.py#L1068-L1084)

## Changes
This patch creates dummy responses for all the skipped items

### Monday
https://quality-match.monday.com/boards/1606259534/pulses/1871935583

### DIFF
--- a/hari_client/upload/hari_uploader.py
+++ b/hari_client/upload/hari_uploader.py
@@ -974,6 +974,7 @@ class HARIUploader:
                         )
                     )
                     failed_media_objects.extend(media.media_objects)
+                    self._media_object_upload_progress.update(len(media.media_objects))
 
                     for media_object in media.media_objects:
                         self.failures.failed_media_object_attributes.extend(
@@ -988,6 +989,9 @@ class HARIUploader:
                             )
                         )
                         failed_media_object_attributes.extend(media_object.attributes)
+                        self._attribute_upload_progress.update(
+                            len(media_object.attributes)
+                        )
 
         # filter out media_objects and attributes that should be skipped because its media failed to upload
         media_objects_to_upload: list[HARIMediaObject] = [
@@ -1050,6 +1054,7 @@ class HARIUploader:
                         )
                     )
                     failed_media_object_attributes.extend(media_object.attributes)
+                    self._attribute_upload_progress.update(len(media_object.attributes))
 
         # update attributes to upload with the media_object attributes that should not be skipped
         for media_object in media_objects_to_upload:
--- a/hari_client/upload/hari_uploader.py
+++ b/hari_client/upload/hari_uploader.py
@@ -1140,21 +1140,25 @@ class HARIUploader:
         """
         # handle skipped media objects and attributes
         if media_objects:
-            skipped_mo_resp = models.BulkResponse()
+            skipped_media_objects_resp = models.BulkResponse()
             for mo in media_objects:
-                skipped_mo_resp.results.append(
+                skipped_media_objects_resp.results.append(
                     models.AnnotatableCreateResponse(
                         bulk_operation_annotatable_id="",
                         item_id=None,
-                        status=models.ResponseStatesEnum.BAD_DATA,
+                        status=models.ResponseStatesEnum.SKIPPED,
                         errors=["Parent media upload failed. Skipping media object."],
                     )
                 )
             # ensure summary matches results count
-            skipped_mo_resp.summary.total = len(skipped_mo_resp.results)
-            skipped_mo_resp.summary.failed = len(skipped_mo_resp.results)
-            skipped_mo_resp.status = models.BulkOperationStatusEnum.FAILURE
-            media_object_upload_responses.append(skipped_mo_resp)
+            skipped_media_objects_resp.summary.total = len(
+                skipped_media_objects_resp.results
+            )
+            skipped_media_objects_resp.summary.failed = len(
+                skipped_media_objects_resp.results
+            )
+            skipped_media_objects_resp.status = models.BulkOperationStatusEnum.FAILURE
+            media_object_upload_responses.append(skipped_media_objects_resp)
 
         if attributes:
             skipped_attr_resp = models.BulkResponse()
@@ -1163,7 +1167,7 @@ class HARIUploader:
                     models.AttributeCreateResponse(
                         annotatable_id="",
                         item_id=None,
-                        status=models.ResponseStatesEnum.BAD_DATA,
+                        status=models.ResponseStatesEnum.SKIPPED,
                         errors=[
                             "Parent media object upload failed. Skipping attribute."
                         ],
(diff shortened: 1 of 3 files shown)


hari_client/upload/hari_uploader.py
```diff
@@ -1114,6 +1122,58 @@ def upload_data_in_batches(self) -> tuple[list[models.BulkResponse], ...]:
...
+
+        Args:
+            media_object_upload_responses: The media object upload responses.
+            media_objects: media objects to generate responses for.
+            attributes_upload_responses: The attribute upload responses.
+            attributes: attributes to generate responses for.
+        """
+        # handle skipped media objects and attributes
+        if media_objects:
+            skipped_mo_resp = models.BulkResponse()
```
katia-qm, line 1143: nitpick: maybe `skipped_media_objects_resp` or `skipped_objects_resp`?

hari_client/upload/hari_uploader.py
```diff
@@ -1114,6 +1122,58 @@ def upload_data_in_batches(self) -> tuple[list[models.BulkResponse], ...]:
...
+            skipped_mo_resp.summary.total = len(skipped_mo_resp.results)
+            skipped_mo_resp.summary.failed = len(skipped_mo_resp.results)
+            skipped_mo_resp.status = models.BulkOperationStatusEnum.FAILURE
+            media_object_upload_responses.append(skipped_mo_resp)
+
+        if attributes:
+            skipped_attr_resp = models.BulkResponse()
+            for attr in attributes:
+                skipped_attr_resp.results.append(
+                    models.AnnotatableCreateResponse(
```
katia-qm, line 1163: should be `AttributeCreateResponse`, and it has a different argument

hari_client/upload/hari_uploader.py
```diff
@@ -1114,6 +1122,58 @@ def upload_data_in_batches(self) -> tuple[list[models.BulkResponse], ...]:
...
+        """
+        # handle skipped media objects and attributes
+        if media_objects:
+            skipped_mo_resp = models.BulkResponse()
+            for mo in media_objects:
+                skipped_mo_resp.results.append(
+                    models.AnnotatableCreateResponse(
+                        bulk_operation_annotatable_id="",
+                        item_id=None,
+                        status=models.ResponseStatesEnum.BAD_DATA,
```
katia-qm, line 1149: should be missing data as in the api

hari_client/upload/hari_uploader.py
```diff
@@ -1114,6 +1122,58 @@ def upload_data_in_batches(self) -> tuple[list[models.BulkResponse], ...]:
...
+            media_object_upload_responses.append(skipped_mo_resp)
+
+        if attributes:
+            skipped_attr_resp = models.BulkResponse()
+            for attr in attributes:
+                skipped_attr_resp.results.append(
+                    models.AnnotatableCreateResponse(
+                        bulk_operation_annotatable_id="",
+                        item_id=None,
+                        status=models.ResponseStatesEnum.BAD_DATA,
```
katia-qm, line 1166: should be missing data as in the api

tests/test_hari_uploader.py
```diff
@@ -1825,7 +1845,25 @@ def mock_create_media_objects(*args, **kwargs):
...
+                    ]
+                ),
+                failed=len(
+                    [
+                        x
+                        for x in results
+                        if x.status != models.ResponseStatesEnum.SUCCESS
+                    ]
+                ),
+            ),
```
katia-qm, line 1866: could you also assert in the test that the response now has right summary equal to the initial number of entities to be uploaded?

hari_client/upload/hari_uploader.py
```diff
@@ -1063,6 +1063,14 @@ def _upload_single_batch(
         attributes_upload_responses = self._upload_attributes_in_batches(
             attributes_to_upload
         )
+        # add fake responses for skipped media objects and attributes
+        self.add_fake_bulk_responses_for_media_objects_and_attributes(
```
katia-qm, line 1067: why did you decide to create all responses all at once? It makes total sense to me, maybe it's better to do it item per item, I just wonder whether we should preserve the items order, probably not

hari_client/upload/hari_uploader.py
Schmaexxi: 21f47411594cf7a04c3f6d69fbc4423879fac94b

tests/test_hari_uploader.py
katia-qm: Let me double check

hari_client/upload/hari_uploader.py
Schmaexxi: About missing data I was mistaken about its use case. Turns out the API would be returning it in this specific case.
I'll update it!

About syncing the response-enum-values, I disagree. You even extended it yourself just now (which I think is fine) https://github.com/quality-match/hari-client/pull/96/files#diff-d2319348921f46d677a0566fa39adc27a82d7736662d5097ebbb59118956ec34R1126
Because it doesn't need to be synced it merely needs to be a superset of the API response enum values

katia-qm:
> Attributes dont have a bulk_..._id-to track local objects after being returned by the server, yet. This, for batches of size > 1, without making assumptions about the order of attributes returned by the server, makes it impossible to match failing attributes back to the locally created attributes Let's do that in another release

They have item_id that is attribute id of the HARIAttribute

Schmaexxi:
> They have item_id that is attribute id of the HARIAttribute

I'm not sure I can follow. The problem would be that during bulk-creation, the server replies with a list of status with notion of success/failure. To match these individual items of the response, for Annotatables, we use the media_bulk_id. For Attributes, this does not exist. An `item_id` or any other set field that would make attributes distinguishable in the response would not be set for failed uploads, making it impossible to match to the item that was tried to upload in a batch.
Did you mean something else?

katia-qm:
> The problem would be that during bulk-creation, the server replies with a list of status with notion of success/failure. To match these individual items of the response, for Annotatables, we use the media_bulk_id. For Attributes, this does not exist. An `item_id` or any other set field that would make attributes distinguishable in the response would not be set for failed uploads, making it impossible to match to the item that was tried to upload in a batch. Did you mean something else?

As I have just checked, we don't send None for failed attributes but their ids as item_id, because id is provided on the creation by the user already, in difference with media and media objects where ids are only created on the back end. (The fact that the user creates an id himself is not a best practice, but it's how it is now and we can add failed response by item_id. Later if we change it it would have to be reworked)

PR #101, opened 2025-06-30
Prepare projection query param for request

katia-qm:
https://quality-match.monday.com/boards/1606259534/pulses/1891324661

Prepare projection query param for request in the needed format

### DIFF
--- a/hari_client/client/client.py
+++ b/hari_client/client/client.py
@@ -186,6 +186,8 @@ def _prepare_request_query_params(
                 + " Support for this behavior will be removed in a future release."
             )
             warnings.warn(msg)
+        elif param_name == "projection" and isinstance(param_value, dict):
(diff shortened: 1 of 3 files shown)


hari_client/client/client.py
```diff
@@ -1023,9 +1027,9 @@ def get_media(
             presign_media: Whether to presign media
             archived: Return archived media
             projection: The fields to be returned (dictionary keys with value True are
-                returned, keys with value False are not returned)
+                returned, keys with value False are not returned). Mixing of True and False values is not allowed.
```
Dens49, line 1030: wow didn't know that. What do you think about adding this as validation to the client?

katia-qm: Makes total sense, added it, could you check again?

PR #102, opened 2025-07-01
Fix json encoding for nested pydantic models

mylius:
[Task](https://quality-match.monday.com/boards/1606259534/pulses/2029064945)

### DIFF
--- a/hari_client/client/client.py
+++ b/hari_client/client/client.py
@@ -492,9 +492,6 @@ class HARIClient:
         Raises:
             APIException: If the request fails.
         """
-        if external_media_source:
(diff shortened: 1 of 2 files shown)


hari_client/client/client.py
```diff
@@ -941,17 +937,15 @@ def create_medias(
                 dataset_id, file_paths=file_paths
             )
 
-            # 2. set media_urls on medias and parse them to dicts
+            # 2. set media_urls on medias
             media_dicts = []
             for idx, media in enumerate(medias):
```
katia-qm, line 942: I guess you have to remove media_dicts and also pass medias to the request instead

PR #103, opened 2025-07-03
Reflect api changes aints

katia-qm:
https://quality-match.monday.com/boards/1606259534/pulses/1891324661
Remove arguments for aint endpoints that were deleter from API models


mylius (approved):
Looks fine to me. I could imagine scenarios where CS complains about not being able to override the ACC, but they probably shouldn't do that.

PR #106, opened 2025-07-14
update bulk size

katia-qm:


hari_client/client/client.py
```diff
@@ -947,9 +947,9 @@ def create_medias(
             MediaFileExtensionNotIdentifiedDuringUploadError: if the file_extension of the provided file_paths couldn't be identified.
         """
 
-        if len(medias) > HARIClient.BULK_UPLOAD_LIMIT:
+        if len(medias) > config.MEDIA_BULK_UPLOAD_LIMIT:
```
Schmaexxi, line 950: this one refers to the endpoint for getting presigned urls, which is set to 2500. We should align them. Setting it to 500 in the BE , again, seems wise.
https://github.com/quality-match/goliat/blob/3ff3362582b79a6146259766a9c7ea5b8d787bcd/hari/hari/backend/routers/medias.py#L67

Schmaexxi: this and all the other occurrences should use the attribute of the HARIUploaderConfig rather than the global variable

hari_client/client/client.py
```diff
@@ -1482,11 +1482,11 @@ def get_presigned_media_upload_url(
             APIException: If the request fails.
             ParameterRangeError: If the validating input args fails.
         """
-        if batch_size < 1 or batch_size > HARIClient.BULK_UPLOAD_LIMIT:
+        if batch_size < 1 or batch_size > HARIClient.MAX_FETCH_BATCH_SIZE:
```
Schmaexxi, line 1485: this limit should be the same as the one used [here](https://github.com/quality-match/hari-client/blob/a0f8fa225968386ec7a202084bafac49ecfcb21b/hari_client/client/client.py#L950) or maybe we default it and make it possible to pass through so they are aligned.
This method is being called from `[create_medias](https://github.com/quality-match/hari-client/blob/a0f8fa225968386ec7a202084bafac49ecfcb21b/hari_client/client/client.py#L950)` where we already check for a limit.
Further, `create_medias` is being called with yet another limit [here](https://github.com/quality-match/hari-client/blob/5a3dab4ee6b4c808b2995a6944120de04d15d657/hari_client/upload/hari_uploader.py#L1108). They could all be unified

Schmaexxi (dismissed):
Thanks for updating the tuning!
Let's make sure we align the correct limits across the different methods they're being used in 👍

hari_client/client/client.py
```diff
@@ -947,9 +945,9 @@ def create_medias(
             MediaFileExtensionNotIdentifiedDuringUploadError: if the file_extension of the provided file_paths couldn't be identified.
         """
 
-        if len(medias) > HARIClient.BULK_UPLOAD_LIMIT:
+        if len(medias) > config.MEDIA_BULK_UPLOAD_LIMIT:
             raise errors.BulkUploadSizeRangeError(
-                limit=HARIClient.BULK_UPLOAD_LIMIT, found_amount=len(medias)
+                limit=config.MEDIA_BULK_UPLOAD_LIMIT, found_amount=len(medias)
```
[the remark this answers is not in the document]
katia-qm: do

Dens49:
would be nice to explain the changes in the changelog

PR #111, opened 2025-07-18
add referenced models

katia-qm:
[task](https://quality-match.monday.com/boards/1606259534/pulses/2056000031)

Referenced models for media and media objects can be used to upload missing objects and attributes for existing media and media objects, based on the back references (just like state aware uploader)


CHANGELOG.md
```diff
@@ -1,24 +1,39 @@
...
+### State aware uploader [PR#96](https://github.com/quality-match/hari-client/pull/96)
+
+#### Breaking Changes
 
 - skip already uploaded media and media objects during upload by default (by comparing existing back references)
 - get existing attribute metadata and reuse the existing ids on new attributes.
 - removed _`media_back_references`, _`media_object_back_references`, _`media_object_cnt` and _`attribute_cnt` fields from the HARIUploader
 
-### New features
+#### New features
```
FBOBecker, line 19: why 4? We usually use 3

Dens49 (changes requested):
One small change request for the changelog and I have a few questions :)
what's the purpose of ReferencedMediaObject and ReferencedMedia?
How does their usage differ from HARIMediaObject and HARIMedia?
When would a user need to use ReferencedMediaObject instead of HARIMediaObject or ReferencedMedia instead of HARIMedia?
If `ReferencedMedia` should refer to media that's already available on the server side, I'd suggest to change the prefix to `Remote*`. Would be a bit clearer imo, but I don't know the goal of the update.

katia-qm:
> Were other alternatives for implementation discussed? For example: adding a new upload method to the uploader, which doesn't rely on HARIAttribute being attached to a HARIMediaObject and that to be attached to HARIMedia.

No, I went for a simple and working solution as the first draft.
Currently uploader only directly works with media, so enabling attaching objects/attributes to the uploader directly would be a major rework. Adding a new upload method to the uploader that possibly accepts plain media objects/ attributes is also the way to do it, but then looking for the right annotatable/media id is up to the user. And if the user does it himself, there is an option for him to just use plain client endpoints for the upload. 
We could of course discuss it further, afaik there was never a use case for that, and the upcoming features for state aware uploader and referenced models should cover all the needs for now imo

PR #118, opened 2025-08-13
Km/clean up duplicates for models

katia-qm:


Dens49 (approved):
thanks for the cleanup! Good choice to mention that as a breaking change :thumbs-up:

